PR #2, opened 2020-02-05
Add url attribute to textfield

iwootten:
### PR Context

Add a new url attribute to textfield to support lookup/typeahead style inputs in runner.

### Checklist

Check you're happy with approach, appropriate tests have been added.


iwootten:
[[C] Add lookup functionality to text field answer type (M)](https://trello.com/c/7ypG3adI/3442-c-add-lookup-functionality-to-text-field-answer-type-m)

tests/schemas/valid/test_text_field_typeahead.json
```diff
@@ -0,0 +1,65 @@
...
+            {
+              "type": "Question",
+              "id": "country-block",
+              "question": {
+                "answers": [
+                  {
+                    "id": "country-answer",
+                    "label": "What is your country of birth?",
+                    "mandatory": false,
+                    "url": "https://gist.githubusercontent.com/rmccar/c123023fa6bd1b137d7f960c3ffa1fed/raw/368a3ea741f72c62c735c319ff7e33e3c1bfdc53/country-of-birth.json",
```
MebinAbraham: I think we should make this URL valid and not commit a personal gist. Maybe just `"https://eq.ons.gov.uk`

iwootten: It is a valid url and doesn't matter what the url is? It's not being visited as part of tests.

MebinAbraham: It's so that we don't need to commit someone personal gist...

app/validation/validator.py
```diff
@@ -2193,6 +2202,12 @@ def _get_block_ids(json_to_validate):
 
         return block_ids
 
+    @staticmethod
+    def _validate_url(url):
+        parsed_result = urlparse(url)
+
+        return parsed_result.scheme and parsed_result.netloc
```
MebinAbraham: A FDQN is enough?. Even without the scheme, the URL would be valid.

tests/schemas/valid/test_text_field_typeahead.json
iwootten: It doesn't matter? It can be any url.

MebinAbraham: The point isn't that it can't be any url, the point is that I don't think it should link to an external gist. May as well make it https://abc.com if it doesn't matter. But okay I fold.

app/validation/validator.py
iwootten: This is to validate a url not a FQDN. It isn't enough because it wouldn't specify the scheme? The design system requires a full url to be able to pull the data.

tests/schemas/invalid/test_text_field_typeahead_invalid_url.json
```diff
@@ -0,0 +1,65 @@
...
+  "data_version": "0.0.3",
+  "survey_id": "001",
+  "title": "Other input fields",
+  "theme": "default",
+  "description": "A questionnaire to test textfields",
+  "messages": {
+    "NUMBER_TOO_LARGE": "Number is too large",
+    "NUMBER_TOO_SMALL": "Number cannot be less than zero",
+    "INVALID_NUMBER": "Please enter an integer"
+  },
```
MebinAbraham: Should remove these redundant data. Same for the other file

tests/schemas/invalid/test_text_field_typeahead_invalid_url.json
```diff
@@ -0,0 +1,65 @@
+{
+  "mime_type": "application/json/ons/eq",
+  "schema_version": "0.0.1",
+  "data_version": "0.0.3",
+  "survey_id": "001",
+  "title": "Other input fields",
```
MebinAbraham: Needs updating.

app/validation/validator.py
```diff
@@ -363,6 +364,16 @@ def _validate_questions(self, block_or_variant, numeric_answer_ranges):
                             self._validate_minimum_and_maximum_offset_date(answer)
                         )
 
+                if answer["type"] == "TextField":
+                    if "suggestions_url" in answer and not self._validate_url(
+                        answer["suggestions_url"]
+                    ):
+                        errors.append(
+                            self._error_message(
+                                f'Answer lookup url used for TextField `{answer["id"]}` is invalid'
```
ajmaddaford: We should be consistent in how we refer to the suggestions url, this should read "Suggestions url used for ..."

tests/schemas/invalid/test_text_field_typeahead_invalid_suggestions_url.json
```diff
@@ -0,0 +1,60 @@
+{
```
ajmaddaford: We should remove 'typeahead' from the filename. Typeahead is an implementation detail of how we're using the suggestions url in runner.

tests/schemas/invalid/test_text_field_typeahead_invalid_suggestions_url.json
```diff
@@ -0,0 +1,60 @@
...
+  "mime_type": "application/json/ons/eq",
+  "schema_version": "0.0.1",
+  "data_version": "0.0.3",
+  "survey_id": "001",
+  "title": "Textfield Typeahead",
```
ajmaddaford: Typeahead -> Suggestions

PR #4, opened 2020-02-12
Ignore labels that have text_plural

iwootten:
### PR Context

Currently validator attempts to check answer labels and values always match.

We shouldn't attempt to validate plurals that are used on answer labels as validator cannot resolve which string is to be used as this is based on the length of a list.


iwootten:
[[B] Implement strings based on a number of items (L)](https://trello.com/c/pKafSvaX/3556-b-implement-strings-based-on-a-number-of-items-l)

PR #6, opened 2020-02-14
Add section summary completion boolean to schema

LJBabbage:
### PR Context
To add a boolean flag into the schema to control the visibility of a section summary. 

N.b Eventual SectionSummary will be removed from block types (summary.json), however this will currently break all the schemas on questionnaire runner and that work has now been included on another story.


tests/schemas/valid/test_section_summary_completion.json
```diff
@@ -0,0 +1,69 @@
+{
```
ajmaddaford: We don't usually create test schemas for new content that is validated by JSON schema. The test schemas test the validation done in code.

PR #8, opened 2020-02-18
Propose custom section summaries

ajmaddaford:
# Context

Currently section summaries don't provide a way to define what should be summarised. To summarise a section that contains list collectors we use information from the first list collector for each list. This forces section summary properties into the first list collector (`add_link_text`, `empty_list_text`) and causes un-necessary complexity when a section summary needs to be rendered.

This pull request proposes schema changes to explicitly define the summary content for sections that contain lists.

### DIFF
--- a/doc/decisions/0006-custom-section-summaries.md
+++ b/doc/decisions/0006-custom-section-summaries.md
@@ -2,7 +2,7 @@
 
 ## Context
 
-Section summaries don't provide a way to define what should be summarised. To summarise a section that contains list collectors we use information from the first list collector for each list. This forces section summary properties into the first list collector (`add_link_text`, `empty_list_text`) and causes un-necessary complexity when a section summary needs to be rendered.
+Section summaries don't provide a way to define what should be summarised. To summarise a section that contains list collectors, we use information from the first list collector for each list. This forces section summary properties into the first list collector (`add_link_text`, `empty_list_text`) and causes un-necessary complexity when a section summary needs to be rendered.
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


doc/decisions/0006-custom-section-summaries.md
```diff
@@ -0,0 +1,69 @@
+# 6. Custom section summaries
+
+## Context
+
+Section summaries don't provide a way to define what should be summarised. To summarise a section that contains list collectors we use information from the first list collector for each list. This forces section summary properties into the first list collector (`add_link_text`, `empty_list_text`) and causes un-necessary complexity when a section summary needs to be rendered.
```
MebinAbraham:
```suggestion
Section summaries don't provide a way to define what should be summarised. To summarise a section that contains list collectors, we use information from the first list collector for each list. This forces section summary properties into the first list collector (`add_link_text`, `empty_list_text`) and causes un-necessary complexity when a section summary needs to be rendered.
```

doc/decisions/0006-custom-section-summaries.md
```diff
@@ -0,0 +1,69 @@
...
+                    "item_title": {}
+                }
+            ]
+        }
+    ]
+}
+```
+
+- `item_title` is a string with placeholders and will work the same as the equivalent list collector property. Duplicating the property in the summary allows for a different representation and will simplify implementations (by not having to search for a definition)
+- The initial implementation will only support a `List` type, but this can be exetended in the future for other summary types
```
MebinAbraham:
```suggestion
- The initial implementation will only support a `List` type, but this can be extended in the future for other summary types
```

doc/decisions/0006-custom-section-summaries.md
```diff
@@ -0,0 +1,69 @@
...
+            ]
+        }
+    ]
+}
+```
+
+## Consequences
+
+- Section summaries for lists are explicit rather than implicit; services that interperet the schema will be simpler. 
+- Properties related to summaries are not present on non summary blocks.
```
MebinAbraham:
```suggestion
- Properties related to summaries are not present on non-summary blocks.
```

PR #19, opened 2020-03-17
Add language property to schema

MebinAbraham:
### PR Context
Added a new required property `language` to schemas.

A schema should define the language it is in. This will also be used by https://github.com/ONSdigital/eq-translations/pull/50

### DIFF
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -10,9 +10,13 @@
     "title",
     "sections",
     "theme",
-    "metadata"
+    "metadata",
+    "language"
   ],
   "properties": {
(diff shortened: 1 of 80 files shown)


schemas/questionnaire_v1.json
```diff
@@ -10,9 +10,13 @@
     "title",
     "sections",
     "theme",
-    "metadata"
+    "metadata",
+    "language"
   ],
   "properties": {
+    "language": {
+      "enum": ["en", "cy", "ga", "eo"]
```
JamesGardiner: Minor point but these should probably be alphabetised

PR #20, opened 2020-03-25
Add contains any

LJBabbage:
### PR Context
As part of the a new schema change "contains any" needs to added to validator. The scenario is as follows:

If 'England or 'Ireland' is chosen but not Other,  then goto block1
if 'England or 'Ireland' is chosen and Other, then goto block 2
if Neither 'England' or 'Ireland' and Other is chosen, then goto block 3

The cleanest way to do this at present is to use 'contains any' on 'England'/'Ireland' which is currently implemented on runner, but isn't included in validator

### DIFF
--- a/schemas/when_rule/compare_to_value.json
+++ b/schemas/when_rule/compare_to_value.json
@@ -15,6 +15,7 @@
         "equals",
         "not equals",
         "contains",
+        "contains any",
         "not contains",
         "greater than",
         "less than",


schemas/when_rule/compare_to_value.json
```diff
@@ -15,6 +15,7 @@
         "equals",
         "not equals",
         "contains",
+        "contains any",
```
MebinAbraham: Not sure why this is needed here?

It is already defined in: https://github.com/ONSdigital/eq-questionnaire-validator/blob/master/schemas/when_rule/compare_to_multiple_values.json#L5-L15

LJBabbage: If you look at the other PR on this story, its falling over because it is missing, but I will take a look.

JamesGardiner: https://github.com/ONSdigital/eq-questionnaire-schemas/pull/28/checks?check_run_id=535935251

MebinAbraham: Yes, but I am confused to why it breaks when it is defined as:
https://github.com/ONSdigital/eq-questionnaire-validator/blob/master/schemas/when_rule/definitions.json#L7-L23

ajmaddaford: `contains any` needs to be used with `values` rather than `value`

PR #22, opened 2020-03-27
Use most recent available python version in pyenv

iwootten:
### PR Context

.python-version was updated to 3.8.2 in order to workaround github actions only using the latest release of 3.8. This isn't available in pyenv so won't install for development environments. We should revert the change to .python-version since this isn't used in github actions anymore and we can't install locally.

### Checklist

Tests all still pass.


JamesGardiner:
Looks like Pyenv has added support for 3.8.2 https://github.com/pyenv/pyenv/releases/tag/v1.2.17

PR #23, opened 2020-03-30
Update yarn dependencies to latest

iwootten:
### PR Context

We are getting security warnings that some of our packages are outdated. This updates all yarn packages to their most recent version.

### Checklist

Tests continue to pass.


MebinAbraham:
Can you update this again. I gets diff lock with latest yarn.

PR #25, opened 2020-04-02
Add summary to question for textfield concatenation

LJBabbage:
### PR Context
Questions should be able to concatenate textfields on the summary page. This allows the element summary in question but also restricts this to just TextFields


schemas/string_interpolation/definitions.json
```diff
@@ -37,6 +37,15 @@
       }
     ]
   },
+  "summary": {
```
MebinAbraham, line 40: I'm not sure why this is part of this file?. This should be a property of the question, hence update the question types to have this new property?.

schemas/string_interpolation/definitions.json
```diff
@@ -37,6 +37,15 @@
...
+    "description": "TextField string concatenations for the summary page",
+    "properties": {
+      "concatenation_type": {
+        "enum": ["NewLine", "Space"]
```
MebinAbraham, line 44: Should be `Newline`?

LJBabbage: Down as NewLine on the story

schemas/string_interpolation/definitions.json
LJBabbage: No idea why I didn't put it in the question definitions, doh

schemas/string_interpolation/definitions.json
MebinAbraham: I think we should also change `NewLine` to `Newline` as a newline character is 1 word.

app/validation/validator.py
```diff
@@ -345,6 +345,13 @@ def _validate_questions(self, block_or_variant, numeric_answer_ranges):
                         )
                     )
 
+                if question.get("summary") and answer["type"] != "TextField":
+                    errors.append(
+                        self._error_message(
+                            f'Summary concatenation can only be used for TextFields, `{answer["id"]}` is invalid'
```
MebinAbraham, line 351: Not a blocked but this could say ``{answer["id"]}` is a `{answer["type"]}``

PR #29, opened 2020-04-20
Remove section summary as block and add section end routing

LJBabbage:
### PR Context
Remove section summary from block type and add new routing rule to allow to route to end of a section


schemas/common_definitions.json
```diff
@@ -271,6 +271,11 @@
               "type": "string",
               "description": "The id of another group. The first block in that group will be routed to."
             },
+            "section": {
+              "type": "string",
+              "description": "Routes to the end of a section",
```
JamesGardiner, line 276:
```suggestion
              "description": "Routes to the end of the current section",
```

Does the suggestion make more sense? The only value allowed is 'end' so I am assuming it is routing within the same section.

schemas/when_rule/definitions.json
```diff
@@ -24,6 +24,10 @@
     },
     "minItems": 1
   },
+  "section": {
+    "type": "string",
+    "enum": ["end"]
```
JamesGardiner, line 29: I think 'summary' would be a more specific value here. `end` is open to interpretation as it could be the last block, or the summary

tests/schemas/invalid/test_invalid_hub_and_spoke_and_summary_confirmation_non_existent.json
```diff
@@ -28,6 +28,7 @@
     {
       "id": "employment-section",
       "title": "Employment",
+      "summary": {},
```
JamesGardiner, line 31: I don't think we should allow empty summary definitions - at a minimum we should force it to be populated, such as we do with the hub.

```json
{
	"summary": {
		"enabled": true
	}
}
```

would do

schemas/when_rule/definitions.json
LJBabbage: Section end can take you to a section summary ,the last block, the summary or the hub. It's being used as a routing_path end flag. Because blocks don't have to have section summaries I would argue that we can't be specific.

tests/schemas/invalid/test_invalid_hub_and_spoke_and_summary_confirmation_non_existent.json
LJBabbage: Completely valid point. I just went with the story description which said they should empty

schemas/common_definitions.json
```diff
@@ -271,6 +271,11 @@
...
+              "enum": ["end"]
```
ajmaddaford, line 277: Our enums are capitalised to differentiate them from other values, this should be `End`.

schemas/when_rule/definitions.json
ajmaddaford: Is this definition used anywhere?

LJBabbage: Ah crap, good spot, I put it in the when, then released that was completely in the wrong place, moved it out to the common definitions and clearly forgot to delete it

tests/schemas/invalid/test_invalid_list_collector_driving_question_multiple_driving_questions.json
```diff
@@ -108,7 +109,7 @@
               "routing_rules": [
                 {
                   "goto": {
-                    "block": "summary",
+                    "section": "end",
```
ajmaddaford, line 112: These need to be changed to `End`

tests/schemas/invalid/test_invalid_hub_and_spoke_and_summary_confirmation_non_existent.json
JamesGardiner: Resolving as this will be followed up in a separate card

PR #30, opened 2020-05-05
Add code to allow question variants with mutually exclusive types

berroar:
### PR Context
This PR has been raised to allow validation of schemas that contain question variants including one of a generic question type with a mutually exclusive question type. This is because our current implementation of Mutually Exclusive questions mean they have to contain multiple answer ids and multiple answers, whereas other types do not.

This issue was highlighted by a request [to add question variants for the Communal Establishment census schema.](https://github.com/ONSdigital/eq-questionnaire-schemas/pull/54)

### Checklist
Check that all tests pass.


berroar:
[[D] CE form - new question variant (S)](https://trello.com/c/h4myjeCB/3837-d-ce-form-new-question-variant-s)

tests/schemas/invalid/test_invalid_inconsistent_default_answers_in_variants.json
```diff
@@ -7,7 +7,7 @@
   "session_timeout_in_seconds": 3,
   "title": "Test Question Variants",
   "theme": "default",
-  "description": "A questionnaire to test default values in quesiton variants",
+  "description": "A questionnaire to test default values in question variants",
```
JamesGardiner, line 10: :+1:

app/validation/validator.py
```diff
@@ -515,6 +515,24 @@ def _get_question_variant_fields_sets(variants):
 
             results["number_of_answers"].add(len(results["answer_ids"]))
 
+        # Code to handle comparison of variants which contain a MutuallyExclusive answer type
+        if (
+            len(results["question_types"]) > 1
+            and "MutuallyExclusive" in results["question_types"]
+        ):
+            results["question_types"].remove("MutuallyExclusive")
```
JamesGardiner, line 523: This will only remove the element at the first index that matches, so if there are more than one MutuallyExclusive answer N>1 won't be removed. Is that the expected behaviour?

berroar: I thought as it is a set there would only ever be one Mutually exclusive element?

JamesGardiner: Ah yeah - I assumed it was a list :+1:

app/validation/validator.py
```diff
@@ -515,6 +515,24 @@ def _get_question_variant_fields_sets(variants):
...
+
+            results["answer_ids"].clear()
+
+            results["number_of_answers"].clear()
+
+            for variant in variants:
+                for answer in variant["question"]["answers"]:
+                    if "-exclusive" not in answer["id"]:
+                        results["answer_ids"].add(answer["id"])
```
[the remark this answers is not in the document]
berroar: Thanks, makes sense

PR #32, opened 2020-05-14
Add show on completion property to summary

MebinAbraham:
### PR Context
Adds the `show_on_completion` property of the summary to determine whether the summary should be should when a section is complete.

### Checklist

### DIFF
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -166,7 +166,7 @@
             "properties": {
               "show_on_completion": {
                 "type": "boolean",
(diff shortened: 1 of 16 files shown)


schemas/questionnaire_v1.json
```diff
@@ -164,6 +164,10 @@
           "summary": {
             "type": "object",
             "properties": {
+              "show_on_completion": {
+                "type": "boolean",
+                "description": "A flag to determine whether to show the section summary on completion of the section."
```
JamesGardiner, line 169:
```suggestion
                "description": "A boolean that determines whether to show the section summary on completion of the section."
```

PR #36, opened 2020-06-01
Add optional warning property to questions

ajmaddaford:
### Context

Adds a new optional `warning` property to questions.

### Checklist


ajmaddaford:
I don't think there is a pattern. There are no valid schemas for `CalculatedSummary` or `DateRange` questions, and `MutuallyExclusive` questions only appear in more involved schemas. 

If want to have valid schemas for all types and properties we would end up duplicating most of the runner test schemas in this repo. I have raised a point on the tech session agenda for a discussion about managing the test schemas between the runner repo and here. For now, I could add the property to the schemas in runner, which will get tested against validator in the runner build.

PR #37, opened 2020-06-08
Add individual_response object

JamesGardiner:
### Checklist

### DIFF
--- a/schemas/individual_response/definitions.json
+++ b/schemas/individual_response/definitions.json
@@ -1,0 +1,18 @@
+{
+  "$id": "https://eq.ons.gov.uk/individual_response/definitions.json",
+  "individual_response": {
+    "type": "object",
+    "properties": {
+      "for_list": {
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+        "description": "The list from which an individual response can be requested"
+      },
+      "individual_section_id": {
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+        "description": "The id of the individual section for which an individual response can be requested"
+      }
+    },
+    "additionalProperties": false,
+    "required": ["for_list", "individual_section_id"]
+  }
+}
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -115,6 +115,9 @@
     "hub": {
       "$ref": "https://eq.ons.gov.uk/hub/definitions.json#/hub"
     },
+    "individual_response": {
+      "$ref": "https://eq.ons.gov.uk/individual_response/definitions.json#/individual_response"
+    },
     "sections": {
       "type": "array",
       "minItems": 1,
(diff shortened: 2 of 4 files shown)


MebinAbraham:
Should update the PR title to match the final changes.

schemas/blocks/interstitial.json
```diff
@@ -20,6 +20,25 @@
...
+            "description": "The list to use when determining whether the individual response definition can be shown"
+          },
+          "individual_section_id": {
+            "type": "string",
+            "description": "The id of the individual section for which an individual response can be requested"
+          }
+        },
+        "additionalProperties": false,
+        "required": ["enabled", "for_list", "individual_section_id"]
+      },
```
MebinAbraham, line 41: We should move this outside interstitial and hub definitions and reference it to avoid the duplication.

schemas/individual_response/definitions.json
```diff
@@ -0,0 +1,22 @@
...
+        "description": "Controls whether the individual response definition can be shown"
+      },
+      "for_list": {
+        "type": "string",
+        "description": "The list to use when determining whether the individual response definition can be shown"
+      },
+      "individual_section_id": {
+        "type": "string",
+        "description": "The id of the individual section for which an individual response can be requested"
+      }
```
MebinAbraham, line 17: Thinking about it, I can't see why we want to keep defining these for each block. These should not be different within 1 questionnaire, therefore these seem like top level properties of the hub and blocks should only needed to specify whether IR is enabled or not.

MebinAbraham, line 17:
```suggestion
      "for_list": {
        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
        "description": "The list to use when determining whether the individual response definition can be shown"
      },
      "individual_section_id": {
        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
        "description": "The id of the individual section for which an individual response can be requested"
      }
```

schemas/individual_response/definitions.json
```diff
@@ -0,0 +1,22 @@
...
+  "individual_response": {
+    "type": "object",
+    "properties": {
+      "enabled": {
+        "type": "boolean",
+        "description": "Controls whether the individual response definition can be shown"
+      },
+      "for_list": {
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+        "description": "The list to use when determining whether the individual response definition can be shown"
```
ajmaddaford, line 12: We shouldn't describe runner implementation specifics in schema descriptions, and this is consistent with the description for the section id:
```suggestion
        "description": "The list from which an individual response can be requested"
```

schemas/individual_response/definitions.json
```diff
@@ -0,0 +1,22 @@
...
+      },
+      "individual_section_id": {
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+        "description": "The id of the individual section for which an individual response can be requested"
+      }
+    },
+    "additionalProperties": false,
+    "required": ["enabled", "for_list", "individual_section_id"]
```
ajmaddaford, line 20: The required list doesn't make sense, if the `enabled` property was set to `false` wouldn't we still have to provide `for_list` and `individual_section_id`?

As the `individual_response` property is optional, we could just drop the enabled property.

PR #43, opened 2020-07-03
Format error paths as pointer strings

iwootten:
### PR Context

Validation completely fails when errors are found in the json schema definition due to error paths being returned as deques and unable to be serialised to json. This PR casts them to strings in order for correct serialisation to occur.

### DIFF
--- a/app/validators/schema_validator.py
+++ b/app/validators/schema_validator.py
@@ -41,7 +41,7 @@ class SchemaValidator(Validator):
             return {}
         except ValidationError as e:
             match = best_match([e])
-            path = "/".join([str(path_element) for path_element in e.path])
+            path = "/".join(str(path_element) for path_element in e.path)
             self.add_error(match.message, verbose=e.message, pointer=f"/{path}")
         except SchemaError as e:
             self.add_error(e)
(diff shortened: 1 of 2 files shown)


iwootten:
[Cast schema error paths to strings](https://trello.com/c/TBMe79QP/3982-cast-schema-error-paths-to-strings)

app/validators/schema_validator.py
```diff
@@ -41,7 +41,7 @@ def validate(self):
             return {}
         except ValidationError as e:
             match = best_match([e])
-            self.add_error(match.message, verbose=e.message, path=e.path)
+            self.add_error(match.message, verbose=e.message, path=str(e.path))
```
MebinAbraham, line 44: I think it's more useful to return a string path like: `sections/2/title"` instead of ` "deque(['sections', 2, 'title'])",`

iwootten: I've made this a pointer, given it's used in translations and easy to covert to.

app/validators/schema_validator.py
```diff
@@ -41,7 +41,8 @@ def validate(self):
             return {}
         except ValidationError as e:
             match = best_match([e])
-            self.add_error(match.message, verbose=e.message, path=e.path)
+            path = "/".join([str(path_element) for path_element in e.path])
```
MebinAbraham, line 44: Minor, but the list isn't required.
```suggestion
            path = "/".join(str(path_element) for path_element in e.path)
```

PR #44, opened 2020-07-06
Add answer default and option value mismatch rule

petechd:
### PR Context
This adds new rule to answer options validator that prevents mismatching answer default and options values. This change was requested on this [Trello card](https://trello.com/c/ohpvGLAQ). It fixes a bug in census schemas where default value differs from answer option value.

### DIFF
--- a/app/validators/answers/option_answer_validator.py
+++ b/app/validators/answers/option_answer_validator.py
@@ -9,7 +9,7 @@ class OptionAnswerValidator(AnswerValidator):
     ANSWER_LABEL_VALUE_MISMATCH = "Found mismatching answer value for label"
     LIST_NAME_MISSING = "List name defined in action params does not exist"
     BLOCK_ID_MISSING = "Block id defined in action params does not exist"
-    DEFAULT_MISMATCH = "Couldn't find matching value for answer default"
+    ANSWER_DEFAULT_MISSING = "Couldn't find matching value for answer default"
 
     def __init__(self, schema_element, questionnaire_schema=None):
         super().__init__(schema_element)
@@ -85,4 +85,6 @@ class OptionAnswerValidator(AnswerValidator):
         if default_value and default_value not in [
             option["value"] for option in self.options
         ]:
-            self.add_error(self.DEFAULT_MISMATCH)
+            self.add_error(
+                self.ANSWER_DEFAULT_MISSING, default=default_value, options=self.options
+            )
--- a/app/validators/answers/option_answer_validator.py
+++ b/app/validators/answers/option_answer_validator.py
@@ -22,7 +22,7 @@ class OptionAnswerValidator(AnswerValidator):
         self.validate_duplicate_options()
         self.validate_labels_and_values_match()
         self.validate_answer_actions()
-        self.validate_default_mismatch()
+        self.validate_default_exists_in_options()
         return self.errors
 
     @cached_property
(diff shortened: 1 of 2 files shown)


petechd:
[Bug: NISRA error on proxy question](https://trello.com/c/ohpvGLAQ/3979-bug-nisra-error-on-proxy-question)

app/validators/answers/option_answer_validator.py
```diff
@@ -77,3 +79,11 @@ def validate_answer_actions(self):
...
             if block_id and block_id not in self.block_ids:
                 self.add_error(self.BLOCK_ID_MISSING, block_id=block_id)
+
+    def validate_default_mismatch(self):
+        if self.answer.get("default") is not None:
+            values = []
+            for option in self.options:
+                values.append(option.get("value"))
+            if self.answer.get("default") not in values:
+                self.add_error(self.DEFAULT_MISMATCH)
```
MebinAbraham, line 89: This could be simplified.
```suggestion
        default_value = self.answer.get("default")
        if default_value and default_value not in [
            option["value"] for option in self.options
        ]:
            self.add_error(self.DEFAULT_MISMATCH)
```

Also, as Rhys mentioned, this will need a unit test.

MebinAbraham: I'm not quite sure on the method naming either. We seem to have a mix of naming it after what it does vs what it should validate against. Maybe the method name should be something like `validate_default_exists_in_options`. Thoughts?

petechd: there was already a variable called `ANSWER_LABEL_VALUE_MISMATCH` and `validate_default_route()` method so I was trying to use a similar name, but your suggestion is much better.

app/validators/answers/option_answer_validator.py
```diff
@@ -9,6 +9,7 @@ class OptionAnswerValidator(AnswerValidator):
     ANSWER_LABEL_VALUE_MISMATCH = "Found mismatching answer value for label"
     LIST_NAME_MISSING = "List name defined in action params does not exist"
     BLOCK_ID_MISSING = "Block id defined in action params does not exist"
+    DEFAULT_MISMATCH = "Couldn't find matching value for answer default"
```
iwootten, line 12:
```suggestion
    DEFAULT_MISMATCH = "Did not find matching value for answer default"
```

app/validators/answers/option_answer_validator.py
```diff
@@ -77,3 +79,10 @@ def validate_answer_actions(self):
 
             if block_id and block_id not in self.block_ids:
                 self.add_error(self.BLOCK_ID_MISSING, block_id=block_id)
+
+    def validate_default_exists_in_options(self):
+        default_value = self.answer.get("default")
+        if default_value and default_value not in [
+            option["value"] for option in self.options
+        ]:
+            self.add_error(self.DEFAULT_MISMATCH)
```
iwootten, line 88: It would be good to make better use of the kwargs argument to add_error to provide better context to the developer. This saves a step of having to hunt it down first.

```python
self.add_error(self.DEFAULT_MISMATCH, default_value=default_value)
```

app/validators/answers/option_answer_validator.py
petechd: Thanks!👍

app/validators/answers/option_answer_validator.py
petechd: Done!👍

PR #47, opened 2020-07-08
Update to only allow single question

petechd:
### PR Context
This removes support for `questions` key in a `section` block from validator and jsonschema. Runner's test schemas has been checked with refactored validator rules and no instances of `questions` were found, except `questions` key in `introduction` block. This change is described on [this Trello card](https://trello.com/c/FvXjNCnc). 
### Checklist


petechd:
[Update validator to only allow single question (S)](https://trello.com/c/FvXjNCnc/3902-update-validator-to-only-allow-single-question-s)

PR #49, opened 2020-07-10
Add interviewer note

petechd:
### PR Context
This adds new rules to jsonschema question and interstitial page definitions to make it compatible with [runner change](https://github.com/ONSdigital/eq-questionnaire-runner/pull/194). The change is described on [this Trello card](https://trello.com/c/rWIoaGMJ). 

### Checklist

### DIFF
--- a/schemas/blocks/definitions.json
+++ b/schemas/blocks/definitions.json
@@ -14,6 +14,9 @@
       },
       "contents": {
         "$ref": "https://eq.ons.gov.uk/common_definitions.json#/contents"
+      },
+      "interviewer_note": {
+        "type": "boolean"
       }
     },
     "additionalProperties": false,
--- a/schemas/blocks/definitions.json
+++ b/schemas/blocks/definitions.json
@@ -16,7 +16,7 @@
         "$ref": "https://eq.ons.gov.uk/common_definitions.json#/contents"
       },
       "interviewer_note": {
-        "type": "boolean"
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/interviewer_note"
       }
     },
     "additionalProperties": false,
--- a/schemas/blocks/definitions.json
+++ b/schemas/blocks/definitions.json
@@ -15,8 +15,8 @@
       "contents": {
         "$ref": "https://eq.ons.gov.uk/common_definitions.json#/contents"
       },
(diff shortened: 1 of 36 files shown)


petechd:
[(I) Interviewer Note highlighting on heading (S)](https://trello.com/c/rWIoaGMJ/3975-i-interviewer-note-highlighting-on-heading-s)

petechd:
> We should have an example of this new field in one of the valid schemas.

We do have one new schema in runner, should I move it to validator too?

schemas/common_definitions.json
```diff
@@ -301,5 +301,9 @@
   "non_empty_string": {
     "type": "string",
     "minLength": 1
+  },
+  "interviewer_note": {
+    "description": "Allows customisation of question and interstitial page title.",
+    "type": "boolean"
```
ajmaddaford, line 307: The intention of the new property was to mark a content or question block as for the "interviewer only", rather than directly associate it with runners implementation. I think we should consider making the property `interviewer_only` and the description something like `Marks the page as for the interviewer only. Only applies for questionnaires that will be completed by an interviewer.`

petechd: Done, thanks!

schemas/blocks/definitions.json
```diff
@@ -14,6 +14,9 @@
       },
       "contents": {
         "$ref": "https://eq.ons.gov.uk/common_definitions.json#/contents"
+      },
+      "interviewer_only": {
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/interviewer_only"
```
MebinAbraham, line 19: Not sure why this is here?. This should only be needed for questions and interstitials not all block types.

petechd: It's only for question and interstitial page, how do I make it interstitial specific if it uses `content` key with all sub definitions in `schemas/blocks/definitions.json`?

MebinAbraham: Should be able to add it to `interstitial.json`

MebinAbraham: As discussed, we should add this to the block level for interstitial/question blocks.

petechd: Moved it to block level, thanks👍

PR #52, opened 2020-07-23
Add warning to hub and rationalise contents

LJBabbage:
### PR Context
The hub no longer needs incomplete/complete keys so they have been removed. The title can now be moved to submission and an extra key has been added in warning

### Checklist
Make sure this is correct with the other repo on the card

### DIFF
--- a/schemas/hub/definitions.json
+++ b/schemas/hub/definitions.json
@@ -34,7 +34,21 @@
             "description": "The warning that will appear on the hub, when the questionnaire is complete."
           }
         },
-        "additionalProperties": false
+        "additionalProperties": false,
+        "anyOf": [
+          {
+            "required": ["button"]
+          },
+          {
+            "required": ["guidance"]
+          },
+          {
+            "required": ["title"]
+          },
+          {
+            "required": ["warning"]
+          }
+        ]
       }
     },
     "required": ["enabled"],
--- a/schemas/hub/definitions.json
+++ b/schemas/hub/definitions.json
@@ -14,37 +14,6 @@
           "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
         }
       },
-      "complete": {
-        "type": "object",
-        "properties": {
-          "title": {
-            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
-            "description": "The title that will appear on the hub, when the questionnaire is complete."
-          },
-          "guidance": {
-            "$ref": "https://eq.ons.gov.uk/hub/definitions.json#/guidance"
-          }
-        },
-        "anyOf": [
-          {
-            "required": ["guidance"]
-          },
-          {
-            "required": ["title"]
-          }
-        ],
-        "additionalProperties": false
-      },
-      "incomplete": {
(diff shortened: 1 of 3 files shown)


schemas/hub/definitions.json
```diff
@@ -14,58 +14,44 @@
...
+          },
           "title": {
             "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
             "description": "The title that will appear on the hub, when the questionnaire is complete."
           },
-          "guidance": {
-            "$ref": "https://eq.ons.gov.uk/hub/definitions.json#/guidance"
+          "warning": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+            "description": "The warning that will appear on the hub, when the questionnaire is complete."
```
ajmaddaford, line 34: The descriptions should be consistent and all use "when the questionnaire is complete" rather than "when the questionnaire is ready to be submitted".

schemas/hub/definitions.json
```diff
@@ -14,58 +14,44 @@
...
           }
         },
+        "additionalProperties": false,
         "anyOf": [
+          {
+            "required": ["button"]
+          },
```
ajmaddaford, line 41: I don't think any of these properties are required, as they either have a default in runner or are optional (guidance).

LJBabbage: My thought was that if you included the submission key it couldn't be empty, it need one child

ajmaddaford: Fair enough, I hadn't thought about it like that.

MebinAbraham: We should use `"minProperties": 1` instead as that is intended for the behaviour we are trying to protect against.

schemas/questionnaire_v1.json
```diff
@@ -115,6 +115,29 @@
...
+          "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+          "description": "The text that appears on the submit button, when the questionnaire is complete."
+        },
+        "guidance": {
+          "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+          "description": "The text guidance that appears, when the questionnaire is complete."
+        },
+        "title": {
+          "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+          "description": "The title that will appear on the hub, when the questionnaire is complete."
```
ajmaddaford, line 131: This shouldn't mention hub, something like main heading or main page title should be used:
```suggestion
          "description": "The main heading when the questionnaire is complete."
```

schemas/questionnaire_v1.json
```diff
@@ -115,6 +115,29 @@
...
+        },
+        "warning": {
+          "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+          "description": "The warning that will appear on the hub, when the questionnaire is complete."
```
ajmaddaford, line 135: Again, this shouldn't mention the hub:
```suggestion
          "description": "The warning text when the questionnaire is complete."
```

PR #57, opened 2020-08-14
Add answer actions for lists

LJBabbage:
### PR Context
add_answer, remove_answer and add_or_edit (primary) have been removed and instead we will use actions in answer options. This PR introduces validation around list actions in answer options and removes the old block way

### Checklist
Check there isn't any other validation we need, and that all the old validation is removed


schemas/answer_actions/definitions.json
```diff
@@ -0,0 +1,43 @@
...
+          "AddOrEditAnswerForListItem"
+        ]
+      }
+    },
+    "additionalProperties": false,
+    "required": ["type"]
+  },
+  "redirectListAnswerActions": {
+    "type": "object",
+    "description": "redirected list answer actions",
```
ajmaddaford, line 21:
```suggestion
    "description": "An action that redirects to a list add question",
```

schemas/blocks/list_collector.json
```diff
@@ -42,12 +42,6 @@
       "for_list": {
         "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
       },
-      "add_answer": {
-        "$ref": "list_collector.json#/answer_reference"
-      },
-      "remove_answer": {
-        "$ref": "list_collector.json#/answer_reference"
-      },
```
ajmaddaford, line 50: It looks like `list_collector.json#/answer_reference` is no longer used anywhere, if so the definition of it should also be removed.

app/validators/blocks/list_collector_validator.py
```diff
@@ -6,27 +6,30 @@
...
-        "present in the answer values"
+
+    ADD_ANSWER_FOR_LIST_ITEM_ACTION_NOT_IN_MAIN_BLOCK = (
+        "AddAnswerForListItem action not found in main block"
     )
-    NON_EXISTENT_LIST_COLLECTOR_REMOVE_ANSWER_VALUE = "The list collector block has a remove_answer_value that is not present in the answer values"
-    REMOVE_ANSWER_REFERENCE_NOT_IN_REMOVE_BLOCK = (
-        "remove_answer reference uses id not found in remove_block"
+
+    REMOVE_ANSWER_FOR_LIST_ITEM_ACTION_NOT_IN_REMOVE_BLOCK = (
```
iwootten, line 24: I think it would be clearer to follow the error name/text used elsewhere in the class for these messages. e.g `NO_ADD_ANSWER_ACTION_FOR_LIST_COLLECTOR = "The list collector main block does not contain an AddAnswerForListItem action"`

app/validators/blocks/primary_person_list_collector_validator.py
```diff
@@ -7,15 +7,13 @@
...
     NO_RADIO_FOR_PRIMARY_PERSON_LIST_COLLECTOR = (
         "The primary person list collector block does not contain a Radio answer type"
     )
-    NON_EXISTENT_PRIMARY_PERSON_LIST_COLLECTOR_ANSWER_VALUE = (
-        "The primary person list collector block has an "
-        "add_or_edit_answer value that is not present in the answer values"
-    )
-    ADD_OR_EDIT_ANSWER_REFERENCE_NOT_IN_MAIN_BLOCK = (
-        "add_or_edit_answer reference uses id not found in main block question"
+    ADD_OR_EDIT_ANSWER_FOR_LIST_ITEM_ACTION_NOT_IN_MAIN_BLOCK = (
```
iwootten, line 15: Same comment here regarding naming/copy.

tests/validators/blocks/test_primary_person_list_collector.py
```diff
@@ -62,20 +42,18 @@ def test_primary_person_invalid_list_collector_non_radio():
...
-        questionnaire_schema,
-    )
+    block = questionnaire_schema.get_block("primary-person-list-collector")
+
+    validator = PrimaryPersonListCollectorValidator(block, questionnaire_schema)
 
     expected_errors = [
         {
-            "message": validator.NON_EXISTENT_PRIMARY_PERSON_LIST_COLLECTOR_ANSWER_VALUE,
+            "message": "AddOrEditAnswerForListItem action not found in main block",
```
iwootten, line 54: This should reference the class variable to prevent having to update copy in multiple places.

app/validators/blocks/list_collector_validator.py
```diff
@@ -35,53 +38,29 @@ class ListCollectorValidator(BlockValidator, ValidateListCollectorQuestionsMixin
...
-                self.block["remove_answer"]["value"],
-                self.NO_RADIO_FOR_LIST_COLLECTOR_REMOVE,
-                self.NON_EXISTENT_LIST_COLLECTOR_REMOVE_ANSWER_VALUE,
-            )
-
-            self.validate_list_collector_answer_ids(self.block)
-            self.validate_other_list_collectors()
-        except KeyError as e:
-            self.add_error(self.LIST_COLLECTOR_KEY_MISSING, key=e)
-        return self.errors
```
iwootten, line 67: Not sure why this has been removed?

app/validators/blocks/list_collector_validator.py
LJBabbage: True, I actually just changed the one that was already there and didn't think much of it.

app/validators/blocks/list_collector_validator.py
LJBabbage: Your right, I was thinking that by removing self.block["add_answer"]["value"] and self.block["remove_answer"]["value"] I couldn't generate a KeyError exception, but I just noticed self.block["remove_block"] which could

LJBabbage:
> What is the difference between `RedirectToAddBlock` and `RedirectToListAddQuestion`?

One has params and comes from a different question and one is internal to the list, between myself and Andrew we came up with the 2 names

app/validators/blocks/list_collector_validator.py
```diff
@@ -6,82 +6,59 @@
...
-        "present in the answer values"
-    )
-    NON_EXISTENT_LIST_COLLECTOR_REMOVE_ANSWER_VALUE = "The list collector block has a remove_answer_value that is not present in the answer values"
-    REMOVE_ANSWER_REFERENCE_NOT_IN_REMOVE_BLOCK = (
-        "remove_answer reference uses id not found in remove_block"
-    )
+
+    NO_REDIRECT_TO_ADD_BLOCK = "RedirectToAddBlock action not found"
+
+    NO_REDIRECT_TO_REMOVE_BLOCK = "RedirectToRemoveBlock action not found"
```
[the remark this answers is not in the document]
LJBabbage: Outdate comment?

MebinAbraham:
> > What is the difference between `RedirectToAddBlock` and `RedirectToListAddQuestion`?
> 
> One has params and comes from a different question and one is internal to the list

Personally I don't think the naming does the differentiation justice. Anyone will always have to look at the definition to remind them one was adding as well as redirecting. Given the action for removing was renamed to `RemoveListItemAndAnswers` to make it more explicit, these actions to me seem like they are `AddListItemAndRedirecToAddBlock` (mouthful so open to suggestions) and `RedirectToListAddQuestion`.

app/validators/blocks/list_collector_validator.py
```diff
@@ -6,82 +6,59 @@
...
-        "present in the answer values"
-    )
-    NON_EXISTENT_LIST_COLLECTOR_REMOVE_ANSWER_VALUE = "The list collector block has a remove_answer_value that is not present in the answer values"
-    REMOVE_ANSWER_REFERENCE_NOT_IN_REMOVE_BLOCK = (
-        "remove_answer reference uses id not found in remove_block"
-    )
+
+    NO_REDIRECT_TO_ADD_BLOCK = "RedirectToAddBlock action not found"
+
+    NO_REMOVE_LIST_ITEM_AND_ANSWERS = "RemoveListItemAndAnswers action not found"
```
MebinAbraham, line 22: Things like these could reference the variable itself. Saves renaming if it were to change.
eg:
```py
    NO_REMOVE_LIST_ITEM_AND_ANSWERS = f"{REMOVE_LIST_ITEM_AND_ANSWERS} action not found"
```

MebinAbraham:
> > > What is the difference between `RedirectToAddBlock` and `RedirectToListAddQuestion`?
> > 
> > 
> > One has params and comes from a different question and one is internal to the list
> 
> Personally I don't think the naming does the differentiation justice. Anyone will always have to look at the definition to remind them one was adding as well as redirecting. Given the action for removing was renamed to `RemoveListItemAndAnswers` to make it more explicit, these actions to me seem like they are `AddListItemAndRedirecToAddBlock` (mouthful so open to suggestions) and `RedirectToListAddQuestion`.

Ignore this, as confused my self on when the add behaviour happend.

app/validators/blocks/block_validator.py
```diff
@@ -62,3 +68,37 @@ def validate_answer_source_reference(self, identifiers):
...
                 self.add_error(self.ANSWER_SELF_REFERENCE, referenced_id=identifier)
+
+    def validate_redirect_to_list_add_block_params(self):
+        collector_questions = self.questionnaire_schema.get_all_questions_for_block(
+            self.block
+        )
+
+        for collector_question in collector_questions:
+            for collector_answer in collector_question["answers"]:
+                for option in collector_answer.get("options", []):
```
MebinAbraham, line 79: The variables should be renamed as they are not just collector questions etc.

app/validators/blocks/list_collector_validator.py
```diff
@@ -6,23 +6,17 @@
 
 class ListCollectorValidator(BlockValidator, ValidateListCollectorQuestionsMixin):
     LIST_COLLECTOR_KEY_MISSING = "Missing key in ListCollector"
-    ADD_ANSWER_REFERENCE_NOT_IN_MAIN_BLOCK = (
-        "add_answer reference uses id not found in main block question"
-    )
+    REDIRECT_TO_LIST_ADD_BLOCK = "RedirectToListAddBlock"
+    REMOVE_LIST_ITEM_AND_ANSWERS = "RemoveListItemAndAnswers"
```
MebinAbraham, line 10: We should add a validation to ensure only `PrimaryPersonListCollector` can use `RemoveListItemAndAnswers` as we only support it in that.

schemas/answer_actions/definitions.json
```diff
@@ -0,0 +1,39 @@
+{
+  "$id": "https://eq.ons.gov.uk/answer_actions/definitions.json",
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "redirect_to_list_add_block": {
+    "type": "object",
+    "description": "An action that redirects to the list add block",
+    "properties": {
+      "type": {
+        "enum": ["RedirectToListAddBlock"]
+      }
```
MebinAbraham, line 10: This should have:
```json
      "params": {
        "type": "object",
        "properties": {
          "block_id": {
          },
          "list_name": {
          }
        },
        "additionalProperties": false,
        "required": ["block_id", "list_name"]
      }
```

schemas/answer_actions/definitions.json
```diff
@@ -0,0 +1,39 @@
...
+          "block_id": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
+          },
+          "list_name": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
+          }
+        },
+        "additionalProperties": false,
+        "required": ["block_id", "list_name"]
+      }
```
MebinAbraham, line 34: This isn't needed given we don't support this outside of a `PrimaryPersonListCollextor`?

app/validators/blocks/block_validator.py
```diff
@@ -62,3 +68,37 @@ def validate_answer_source_reference(self, identifiers):
...
+                        ):
+                            self.add_error(self.LIST_NAME_MISSING, list_name=list_name)
+
+                        block_id = params.get("block_id")
+
+                        if (
+                            block_id
+                            and block_id not in self.questionnaire_schema.block_ids
+                        ):
+                            self.add_error(self.BLOCK_ID_MISSING, block_id=block_id)
```
MebinAbraham, line 104: Once the comment about updating the JSON schema is done, this can be simplfied to:
```py
if list_name := params["list_name"] not in self.questionnaire_schema.list_names:
    self.add_error(self.LIST_NAME_MISSING, list_name=list_name)

if block_id := params["block_id"] not in self.questionnaire_schema.block_ids:
    self.add_error(self.BLOCK_ID_MISSING, block_id=block_id)
```
However, the walrus operator means you will need to update black formatting to latest version so, you could do `list_name = params["list_name"]` for now if you wanted to keep updating the package separate.

MebinAbraham (changes requested):

app/validators/blocks/block_validator.py
```diff
@@ -7,6 +7,9 @@ class BlockValidator(Validator):
     ANSWER_REFERENCE_INVALID = "Invalid answer reference"
     LIST_REFERENCE_INVALID = "Invalid list reference"
     ANSWER_SELF_REFERENCE = "Invalid answer reference (self-reference)"
+    LIST_NAME_MISSING = "List name defined in action params does not exist"
+    BLOCK_ID_MISSING = "Block id defined in action params does not exist"
```
ajmaddaford, line 11: I think `MISSING` implies the property has been omitted (which is covered by the schema), these two errors would be clearer if `MISSING` was replaced with `DOESNT_EXIST`.

app/validators/blocks/block_validator.py
```diff
@@ -7,6 +7,9 @@ class BlockValidator(Validator):
...
+    ACTION_PARAMS_MISSING = "Action params does not exist"
```
ajmaddaford, line 12: This message should be consistent with `LIST_COLLECTOR_KEY_MISSING`, something like:
```suggestion
    ACTION_PARAMS_MISSING = "Action params key missing"
```

app/validators/blocks/block_validator.py
```diff
@@ -22,6 +25,9 @@ def validate(self):
 
         self.validate_source_references(source_references)
 
+        if self.block["type"] not in ["ListCollector", "PrimaryPersonListCollector"]:
+            self.validate_redirect_to_list_add_block_params()
```
ajmaddaford, line 29: We should be checking both sides of this - that the params exist for non-collector blocks, and that they don't exist for collector blocks.

app/validators/blocks/block_validator.py
```diff
@@ -62,3 +68,37 @@ def validate_answer_source_reference(self, identifiers):
                 == self.block["id"]
             ):
                 self.add_error(self.ANSWER_SELF_REFERENCE, referenced_id=identifier)
+
+    def validate_redirect_to_list_add_block_params(self):
+        collector_questions = self.questionnaire_schema.get_all_questions_for_block(
+            self.block
+        )
```
ajmaddaford, line 75: These are not collector questions as this method is only called for non-collector blocks.

app/validators/blocks/block_validator.py
```diff
@@ -62,3 +68,37 @@ def validate_answer_source_reference(self, identifiers):
...
+
+        for collector_question in collector_questions:
+            for collector_answer in collector_question["answers"]:
+                for option in collector_answer.get("options", []):
+                    action = option.get("action")
+
```
ajmaddaford, line 81: The validation after this point should only be done if the action type is `RedirectToListAddBlock`.

MebinAbraham (commented):
There are certain edge cases like, atm you can define `RedirectToListAddBlock` in `ListRemoveQuestion` blocks which probably shouldn't be allowed but it is probably over the top in terms of validation.

app/validators/blocks/block_validator.py
```diff
@@ -62,3 +68,37 @@ def validate_answer_source_reference(self, identifiers):
...
+                        if not params:
+                            self.add_error(
+                                self.ACTION_PARAMS_MISSING, block_id=self.block
+                            )
+                            continue
+
+                        list_name = params.get("list_name")
+
+                        if (
+                            list_name
```
ajmaddaford, line 93: If the action type is checked above, then we won't need to check for list name as we've already checked for params and the schema enforces the existence of the `list_name` property.

app/validators/blocks/block_validator.py
```diff
@@ -62,3 +68,37 @@ def validate_answer_source_reference(self, identifiers):
...
+                            and list_name not in self.questionnaire_schema.list_names
+                        ):
+                            self.add_error(self.LIST_NAME_MISSING, list_name=list_name)
+
+                        block_id = params.get("block_id")
+
+                        if (
+                            block_id
```
ajmaddaford, line 101: Same as the comment above, if we check the action type then we don't need to check for the existence of the `block_id` property.

app/validators/blocks/list_collector_validator.py
ajmaddaford, line 10:
```suggestion
    REDIRECT_TO_LIST_ADD_BLOCK_ACTION = "RedirectToListAddBlock"
    REMOVE_LIST_ITEM_AND_ANSWERS_ACTION = "RemoveListItemAndAnswers"
```

app/validators/blocks/primary_person_list_collector_validator.py
```diff
@@ -7,16 +7,11 @@
 class PrimaryPersonListCollectorValidator(
     BlockValidator, ValidateListCollectorQuestionsMixin
 ):
+    REDIRECT_TO_LIST_ADD_BLOCK = "RedirectToListAddBlock"
+    NO_REDIRECT_TO_LIST_ADD_BLOCK = f"{REDIRECT_TO_LIST_ADD_BLOCK} action not found"
```
ajmaddaford, line 11:
```suggestion
    REDIRECT_TO_LIST_ADD_BLOCK_ACTION = "RedirectToListAddBlock"
    NO_REDIRECT_TO_LIST_ADD_BLOCK = f"{REDIRECT_TO_LIST_ADD_BLOCK_ACTION} action not found"
```

schemas/answer_actions/definitions.json
```diff
@@ -0,0 +1,39 @@
...
+      "type": {
+        "enum": ["RedirectToListAddBlock"]
+      }
+    },
+    "additionalProperties": true,
+    "required": ["type"]
+  },
+  "remove_list_item_and_answers": {
+    "type": "object",
+    "description": "An action that redirects to the remove block",
```
ajmaddaford, line 17:
```suggestion
    "description": "An action that removes a list item and all answers associated with it",
```

schemas/answer_actions/definitions.json
ajmaddaford, line 34: It looks like this should be in the other action.

ajmaddaford (changes requested):

MebinAbraham (approved):

ajmaddaford (approved):

PR #58, opened 2020-08-27
Add email confirmation

LJBabbage:
### PR Context
Add email confirmation to schema


schemas/questionnaire_v1.json
```diff
@@ -133,6 +133,10 @@
         "warning": {
           "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
           "description": "The warning when the questionnaire is complete."
+        },
+        "email_confirmation": {
+          "type": "boolean",
+          "description": "A boolean that is used to determine if an email confirmation form is available on questionnaire completion"
```
MebinAbraham, line 139: - `A flag that is used to determine if an email confirmation is available on questionnaire completion`
or just 
- `Used to determine if an email confirmation is available on questionnaire completion`
Don't think the `A boolean` is needed given the type is already a bool.

MebinAbraham (approved):

JamesGardiner (approved):

PR #59, opened 2020-08-27
Add support for address answers

MebinAbraham:
### PR Context
Add support for new "Address" answer type introduced in: https://github.com/ONSdigital/eq-questionnaire-runner/pull/261

### Checklist

### DIFF
--- a/schemas/answers/address.json
+++ b/schemas/answers/address.json
@@ -1,0 +1,37 @@
+{
+  "$id": "https://eq.ons.gov.uk/answers/address.json",
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "answer": {
+    "type": "object",
+    "properties": {
+      "id": {
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/identifier"
+      },
+      "q_code": {
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/q_code"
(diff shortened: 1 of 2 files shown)


schemas/answers/address.json
```diff
@@ -0,0 +1,37 @@
+{
+  "$id": "https://eq.ons.gov.uk/answers/address.json",
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "answer": {
+    "type": "object",
+    "properties": {
+      "id": {
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/identifier"
+      },
+      "q_code": {
```
LJBabbage, line 10: q_code? Are we supporting business on this one

MebinAbraham: Runner still support data version 0.0.1 hence `q_code` is kept.

berroar (commented):
Does this need a valid test schema to go with it?

MebinAbraham:
> Does this need a valid test schema to go with it?

I'm not sure we need it for new JSON definitions as we don't currently do this for example all answers/question types. We tend to when we have custom python logic to go alongside it. But I guess we have had mixed solutions.

berroar (approved):

LJBabbage (approved):

PR #61, opened 2020-09-07
Add page_title definitions

JamesGardiner:
### PR Context
- Updates decision 0007 to take into account non-block page titles and adds additional examples
- Adds the page title property to relevant definitions

### Checklist

	- https://github.com/ONSdigital/eq-translations/pull/64

### DIFF
--- a/schemas/blocks/interstitial.json
+++ b/schemas/blocks/interstitial.json
@@ -12,7 +12,6 @@
       },
       "page_title": {
         "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
-        "description": "Custom page title used on the interstitial page."
       },
       "type": {
         "type": "string",
--- a/schemas/blocks/interstitial.json
+++ b/schemas/blocks/interstitial.json
@@ -12,7 +12,7 @@
       },
       "page_title": {
         "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
-        "description": "Custom page title used the interstitial page."
+        "description": "Custom page title used on the interstitial page."
       },
       "type": {
         "type": "string",
--- a/schemas/blocks/interstitial.json
+++ b/schemas/blocks/interstitial.json
@@ -11,7 +11,7 @@
         "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
       },
       "page_title": {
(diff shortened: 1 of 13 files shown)


schemas/questionnaire_v1.json
```diff
@@ -155,6 +155,10 @@
             "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
             "description": "Title used for the navigation container."
           },
+          "page_title": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+            "description": "Custom Page Title used for section summary pages."
+          },
```
ajmaddaford, line 161: I think this would be better placed in under the `summary` property.

schemas/blocks/interstitial.json
```diff
@@ -10,6 +10,10 @@
       "number": {
         "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
       },
+      "page_title": {
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+        "description": "Custom page title used on the interstitial page."
+      },
```
ajmaddaford, line 16: I think it would be better to make all uses of `page_title` use a reference with a generic description. The context of what page the property is used on is covered by it being in that block types definition.

ajmaddaford (changes requested):

schemas/questionnaire_v1.json
JamesGardiner: I went back and fore on this - happy to go with the above suggestion though.

schemas/blocks/interstitial.json
```diff
@@ -10,6 +10,9 @@
       "number": {
         "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
       },
+      "page_title": {
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
+      },
```
MebinAbraham, line 15: Any reason this is not added to all block types?.

JamesGardiner: No - and it probably should be

JamesGardiner: Done

MebinAbraham (approved):

ajmaddaford (approved):

PR #68, opened 2020-09-23
Add support to placeholders for composite answer values

LJBabbage:
### PR Context
Add support to placeholders for composite answer values

N.B The proposed structure in the card was as follows

```
{
  "answer_id": "address_id",
  "value": {
    "line1": "7 Evelyn Street",
    "town": "Barry",
    "postcode": "CF63 4EN"
  }
}

{ 
    "source": "answers",
    "identifier": "address_id",
    "selector": "line1"
}

```


schemas/string_interpolation/definitions.json
```diff
@@ -173,6 +173,11 @@
         "type": "string",
         "enum": ["same_name_items"]
       },
+      "selector": {
+        "type": "string",
+        "description": "Address answer type field selector",
+        "enum": ["line1", "line2", "town", "postcode"]
+      },
```
MebinAbraham, line 180: I am a bit torn on this as the idea was to add support for composite answer values not just addresses. Given you already check specifically the address answer type in code, the "selectors" that can be used could live in code which will make this more flexible.

app/validators/blocks/block_validator.py
```diff
@@ -56,16 +59,23 @@ def validate_list_source_reference(self, identifiers):
...
+            elif answers_with_context[identifier]["block"] == self.block["id"]:
                 self.add_error(self.ANSWER_SELF_REFERENCE, referenced_id=identifier)
 
+            if (
+                has_selector
+                and answers_with_context[identifier]["answer"]["type"] != "Address"
+            ):
+                self.add_error(
+                    self.SELECTOR_USED_FOR_NON_ADDRESS_TYPE, referenced_id=identifier
+                )
```
MebinAbraham, line 77: This enforces that "selector" can only be used for addresses but I think we should update this to something like `if has_selector and identifier not in ANSWER_TYPES_WITH_COMPOSITE_VALUES` where this can just be `ANSWER_TYPES_WITH_COMPOSITE_VALUES = {"Address"}` for now.
Thoughts?

MebinAbraham (changes requested):
> I have implemented the code like this, but I do have reservations over the name "selector" being too generic and confusing as its only used for Address type.

> This has been highlighted even further, as on merge the id_selector has now appeared, which now makes this far too confusing

Any suggestion to an alternative to "selector"?. We do answer have "id_selector" which could be confusing. Anyhow, I don't think the key shouldn't be specific to an answer type as we should be adding support for composite answer values types even if that is only 1 answer type right now. That is how we have designed placeholders in general.

MebinAbraham:
Just raising this for visibility as I am not sure if we considered it. For our current requirement it is fine but in future we may want to support multiple selectors.

LJBabbage:
I think we might need to take this into tech session, and confirm how we want to take this forward. Originally it was purely for addresses and validate as such (i.e line1, line2, town, postcode) The point regarding multiple selectors is also one to consider

ajmaddaford:
I thought the intent of this work was as Mebin has described - the card is titled "Add support to placeholders for composite answer values", and there is nothing on the card about restricting it to only address answers.

LJBabbage:
> I thought the intent of this work was as Mebin has described - the card is titled "Add support to placeholders for composite answer values", and there is nothing on the card about restricting it to only address answers.

There seems to be some confusion over a chat on the 18th

app/validators/blocks/block_validator.py
```diff
@@ -36,7 +38,8 @@ def validate_source_references(self, source_references):
                 identifiers = source_reference["identifier"]
 
             if source == "answers":
-                self.validate_answer_source_reference(identifiers)
+                has_selector = "selector" in source_reference
+                self.validate_answer_source_reference(identifiers, has_selector)
```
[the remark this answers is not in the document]
LJBabbage: Made awkward due to the lines directly above which make identifiers into a list if not already one, which the other method also use

app/validators/blocks/block_validator.py
```diff
@@ -3,12 +3,16 @@
...
     ACTION_PARAMS_MISSING = "Action params key missing"
     ACTION_PARAMS_SHOULDNT_EXIST = "Action params key should not exist"
+    ANSWER_REFERENCE_INVALID = "Invalid answer reference"
+    ANSWER_SELF_REFERENCE = "Invalid answer reference (self-reference)"
+    COMPOSITE_ANSWER_INVALID = "Invalid composite answer"
+    COMPOSITE_ANSWER_FIELD_INVALID = "Invalid field for composite answer"
+    LIST_REFERENCE_INVALID = "Invalid list reference"
+    METADATA_REFERENCE_INVALID = "Invalid metadata reference"
+
+    COMPOSITE_ANSWERS = {"Address": ["line1", "line2", "town", "postcode"]}
```
MebinAbraham, line 15: Name implies this is a list etc maybe `COMPOSITE_ANSWERS_TO_SELECTORS_MAP` to be explicit?

MebinAbraham (approved):
PR title should be updated.

app/validators/blocks/block_validator.py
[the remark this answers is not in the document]
MebinAbraham: I would say both the methods could just pass the source_reference instead of the identifiers, but not a blocker for this.

ajmaddaford (approved):

PR #69, opened 2020-09-23
Add mobile number answer type

iwootten:
### PR Context

Adds a 'mobile_number' answer type to validator to support the validation of mobile numbers in eq-questionnaire-runner.

### DIFF
--- a/tests/schemas/valid/test_mobile_number.json
+++ b/tests/schemas/valid/test_mobile_number.json
@@ -1,57 +1,57 @@
 {
-    "mime_type": "application/json/ons/eq",
-    "language": "en",
-    "schema_version": "0.0.1",
-    "data_version": "0.0.3",
-    "survey_id": "001",
-    "title": "Test mobile number",
-    "theme": "default",
-    "description": "A questionnaire to test a mobile number",
-    "metadata": [
+  "mime_type": "application/json/ons/eq",
+  "language": "en",
+  "schema_version": "0.0.1",
+  "data_version": "0.0.3",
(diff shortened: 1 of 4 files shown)


iwootten:
[[H] Runner to Validate mobile telephone number answers (M)](https://trello.com/c/vtg2cAV1/4110-h-runner-to-validate-mobile-telephone-number-answers-m)

tests/schemas/valid/test_mobile_number.json
```diff
@@ -0,0 +1,58 @@
+{
+  "mime_type": "application/json/ons/eq",
+  "language": "en",
+  "schema_version": "0.0.1",
+  "data_version": "0.0.3",
+  "survey_id": "001",
+  "title": "Other input fields",
```
MebinAbraham, line 7: I know this is only in validator but name should be relevant. Also, we should try to avoid unnecessary properties in these schemas like `q_code` which is only relevant to data version 0.0.1 anyway.

MebinAbraham (approved):

tests/schemas/valid/test_mobile_number.json
iwootten: Sure, these have been copied from another schema. Would be good if we had some sort of authoring tool for this 🤔

MebinAbraham: Ya, I think is an issue in a lot of schemas unfortunately.

ajmaddaford (approved):

PR #71, opened 2020-10-07
Add initial support for unrelated relationship block

MebinAbraham:
### PR Context
Adds support for the unrelated relationships block implement in https://github.com/ONSdigital/eq-questionnaire-runner/pull/339

To keep relationships consistent, I have enforced `RelationshipCollector` to have an ID of `relationships`. This makes grouping relationship routes in runner easier and avoid hardcoding the ID for the unrelated block.

Open to suggestions in terms of naming the new block.

### Checklist

### DIFF
--- a/schemas/blocks/relationship_collector.json
+++ b/schemas/blocks/relationship_collector.json
@@ -27,7 +27,7 @@
       "question_variants": {
         "$ref": "definitions.json#/question_variants"
       },
-      "unrelated_relationship_block": {
+      "unrelated_block": {
         "$ref": "https://eq.ons.gov.uk/blocks/unrelated_relationship_block.json#/block"
       },
       "routing_rules": {
--- a/schemas/blocks/relationship_collector.json
+++ b/schemas/blocks/relationship_collector.json
@@ -28,7 +28,7 @@
         "$ref": "definitions.json#/question_variants"
       },
       "unrelated_block": {
-        "$ref": "https://eq.ons.gov.uk/blocks/unrelated_relationship_block.json#/block"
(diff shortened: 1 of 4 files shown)


schemas/blocks/unrelated_relationship_block.json
```diff
@@ -0,0 +1,41 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "block": {
+    "type": "object",
+    "properties": {
+      "id": {
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/identifier"
+      },
+      "type": {
+        "type": "string",
+        "enum": ["UnrelatedRelationship"]
```
ajmaddaford, line 12: As distinct from related relationships? I think `UnrelatedQuestion` is better, and more consistent with the list question sub blocks.

schemas/blocks/relationship_collector.json
```diff
@@ -26,6 +27,9 @@
       "question_variants": {
         "$ref": "definitions.json#/question_variants"
       },
+      "unrelated_relationship_block": {
```
ajmaddaford, line 30: I think `unrelated_block` is better. I know there was some concern over it being mis-interpreted as a block that has nothing to do with the parent block, but if that was the case it wouldn't be here 🙂 .

schemas/blocks/unrelated_relationship_block.json
```diff
@@ -0,0 +1,41 @@
+{
```
ajmaddaford, line 1: The filename should be `unrelated_question.json`.

LJBabbage (approved):

ajmaddaford (approved):

PR #72, opened 2020-10-08
Update python version to 3.9.0

petechd:

This updates schemas python version to 3.9.0. Github actions and dockerfile are also updated.

### How to review

Check if tests and Makefile scripts are working fine. Run local copy of validator against local schemas.


berroar:
[Python 3.9 (L)](https://trello.com/c/sGLihMaz/4198-python-39-l)

.github/workflows/pull_request.yml
```diff
@@ -10,9 +10,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-python@v1
+      - uses: actions/setup-python@v2
         with:
-          python-version: 3.8
+          python-version: 3.9
```
berroar, line 15: We should update this action to use a variable here (e.g. ${{ env.PYTHON_VERSION }}) rather than hardcoding to make it consistent with the other repos.

.github/workflows/pull_request.yml
```diff
@@ -53,9 +53,9 @@ jobs:
...
```
berroar, line 58: Same here

petechd: Done, thanks!

.github/workflows/pull_request.yml
petechd: Done, thanks!

MebinAbraham (approved):

JamesGardiner (approved):

PR #79, opened 2020-11-02
Add checkbox answer field no label definition

petechd:
### PR Context
This adds new "no_label" Boolean field to checkbox answer block definitions. It is required by this [card](https://trello.com/c/91PIVgpv) and associated with [this change](https://github.com/ONSdigital/eq-questionnaire-runner/pull/383) in runner and [schemas PR](https://github.com/ONSdigital/eq-questionnaire-schemas/pull/193).

### Checklist

### DIFF
--- a/app/validators/questionnaire_validator.py
+++ b/app/validators/questionnaire_validator.py
@@ -88,7 +88,7 @@ class QuestionnaireValidator(Validator):
 
             for schema_text in values_to_check:
                 if schema_text and quote_regex.search(schema_text):
-                        self.add_error(
-                            error_messages.DUMB_QUOTES_FOUND,
-                            pointer=translatable_item.pointer,
-                        )
+                    self.add_error(
+                        error_messages.DUMB_QUOTES_FOUND,
+                        pointer=translatable_item.pointer,
+                    )
--- a/app/validators/questionnaire_validator.py
+++ b/app/validators/questionnaire_validator.py
@@ -87,10 +87,7 @@ class QuestionnaireValidator(Validator):
                 values_to_check = schema_text.values()
 
             for schema_text in values_to_check:
-                if schema_text is not None:
-                    found = quote_regex.search(schema_text)
-
-                    if found:
+                if schema_text and quote_regex.search(schema_text):
                         self.add_error(
                             error_messages.DUMB_QUOTES_FOUND,
                             pointer=translatable_item.pointer,
--- a/app/validators/questionnaire_validator.py
+++ b/app/validators/questionnaire_validator.py
@@ -87,10 +87,11 @@ class QuestionnaireValidator(Validator):
                 values_to_check = schema_text.values()
 
             for schema_text in values_to_check:
-                found = quote_regex.search(schema_text)
+                if schema_text is not None:
+                    found = quote_regex.search(schema_text)
 
-                if found:
-                    self.add_error(
-                        error_messages.DUMB_QUOTES_FOUND,
-                        pointer=translatable_item.pointer,
-                    )
+                    if found:
+                        self.add_error(
+                            error_messages.DUMB_QUOTES_FOUND,
+                            pointer=translatable_item.pointer,
+                        )
--- a/schemas/answers/checkbox.json
+++ b/schemas/answers/checkbox.json
@@ -55,16 +55,12 @@
     "dependencies": {
       "label": {
         "not": {
-          "required": [
-            "no_label"
-          ]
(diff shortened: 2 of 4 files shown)


petechd:
[P1: Checkbox default text to be removed - NI](https://trello.com/c/91PIVgpv/4266-p1-checkbox-default-text-to-be-removed-ni)

schemas/answers/checkbox.json
```diff
@@ -36,6 +36,9 @@
       "mandatory": {
         "type": "boolean"
       },
+      "no_label": {
```
LJBabbage, line 39: Think we need to make no_label exclusive to label. At the moment it's possible to have a label and a no_label flag, which doesn't make sense

LJBabbage (changes requested):

schemas/answers/checkbox.json
petechd: Sorted using dependency, thanks!

LJBabbage (changes requested):
Nice use of dependencies, not seen this before, I do reckon we need a invalid test for this one, e.g. a test that has a checkbox with both label and no_label

schemas/answers/checkbox.json
```diff
@@ -13,6 +13,9 @@
       "label": {
         "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
       },
+      "no_label": {
+        "type": "boolean"
+      },
```
MebinAbraham, line 18: If the only way we are going to solve this for now is with a new property then I think it would be better to have something like `show_label` which can be default to `True` in runner. The `"no_label": false` double negative isn't that great. Might be worth discussing with the rest of the team, have discussed with Lewis and I believe he is on board. I don't think there is a perfect solution but think "show" is clearer.

petechd:
> Nice use of dependencies, not seen this before, I do reckon we need a invalid test for this one, e.g. a test that has a checkbox with both label and no_label

Added invalid test schema with both `label` and `show_label`.

schemas/answers/checkbox.json
petechd: Added `show_label` now set to `False` if label shouldn't be visible. Thanks!

LJBabbage:
> Added invalid test schema with both `label` and `show_label`.

The invalid test schema on it's own won't actually do anything, you need a test to exercise it

ajmaddaford:
I think the new property is confusing. We should consider allowing `null` for label instead, so in runner:
- no label property would display the default label
- a label property set to `null` would not display the label

petechd:
> I think the new property is confusing. We should consider allowing `null` for label instead, so in runner:
> 
> * no label property would display the default label
> * a label property set to `null` would not display the label

I've added `null` value to label so when it's used it will not display any label, no label will display default value. Thanks!

app/validators/questionnaire_validator.py
```diff
@@ -87,10 +87,11 @@ def validate_smart_quotes(self):
...
 
             for schema_text in values_to_check:
-                found = quote_regex.search(schema_text)
-
-                if found:
-                    self.add_error(
-                        error_messages.DUMB_QUOTES_FOUND,
-                        pointer=translatable_item.pointer,
-                    )
+                if schema_text is not None:
```
ajmaddaford, line 90: I think this might highlight an underlying issue with a `None` value being in `translatable_items`. I think we should test this with translations to see if it breaks it - it could error trying to lookup a translation value of `None`.

petechd: I tested translation templates and it looks like `None` labels are not being added to pot files, so it looks like it doesn't break translations.

ajmaddaford: Did you test generating a translated schema as well as extraction?

petechd: I did both and it works fine, `null` labels are not added to .pot files and generating schemas from .pot files is working as expected, I created automated test for eq-translations, should I add it?

app/validators/questionnaire_validator.py
```diff
@@ -87,10 +87,11 @@ def validate_smart_quotes(self):
...
+                    found = quote_regex.search(schema_text)
+
+                    if found:
```
ajmaddaford, line 93: I think this can be simplified to:
```suggestion
                if schema_text and quote_regex.search(schema_text):
```

ajmaddaford (approved):

LJBabbage (approved):

PR #81, opened 2020-11-06
Add actions validation for unrelated blocks

ajmaddaford:
### PR Context

Adds validation for the required actions (`RemoveUnrelatedRelationships` and `AddUnrelatedRelationships`) on unrelated blocks.

### Checklist

### DIFF
--- a/app/validators/blocks/relationship_collector_validator.py
+++ b/app/validators/blocks/relationship_collector_validator.py
@@ -1,5 +1,6 @@
 from app import error_messages
 from app.validators.blocks.block_validator import BlockValidator
+from app.validators.blocks.unrelated_block_validator import UnrelatedBlockValidator
 
 
 class RelationshipCollectorValidator(BlockValidator):
@@ -16,6 +17,8 @@ class RelationshipCollectorValidator(BlockValidator):
         self.validate_list_exists()
         self.validate_multiple_answers()
         self.validate_answer_type()
+        self.validate_unrelated_block()
+
         return self.errors
 
     def validate_answer_type(self):
@@ -38,3 +41,10 @@ class RelationshipCollectorValidator(BlockValidator):
 
         if list_name not in self.questionnaire_schema.list_names:
             self.add_error(error_messages.FOR_LIST_NEVER_POPULATED, list_name=list_name)
+
+    def validate_unrelated_block(self):
+        if "unrelated_block" in self.block:
+            block_validator = UnrelatedBlockValidator(
+                self.block["unrelated_block"], self.questionnaire_schema
+            )
+            self.errors += block_validator.validate()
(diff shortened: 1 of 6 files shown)


app/validators/blocks/relationship_collector_validator.py
```diff
@@ -1,5 +1,6 @@
 from app import error_messages
 from app.validators.blocks.block_validator import BlockValidator
+from app.validators.blocks.unrelated_block_validator import UnrelatedBlockValidator
```
JamesGardiner, line 3: A minor niggle, but these imports could be:

```
from app.validators import BlockValidator
from app.validators import UnrelatedBlockValidator
```

with some small changes to the init modules

JamesGardiner (approved):
LGTM - minor suggestion to improve readability of imports, but it's not blocking

app/validators/blocks/relationship_collector_validator.py
ajmaddaford: I agree, but this is consistent with the imports throughout validator. This change should be made across all files in a separate pull request.

app/validators/blocks/unrelated_block_validator.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def validate_answer_actions(self):
+        expected_actions = ["AddUnrelatedRelationships", "RemoveUnrelatedRelationships"]
+        questions = self.questionnaire_schema.get_all_questions_for_block(self.block)
+        for question in questions:
+            actions = []
+            for answer in question["answers"]:
+                for option in answer["options"]:
+                    if "action" in option:
+                        actions.append(option["action"]["type"])
```
MebinAbraham, line 20: Ideal for a comprehension?
```suggestion
            actions = [
                option["action"]["type"]
                for answer in question["answers"]
                for option in answer["options"]
                if "action" in option
            ]
```

app/validators/blocks/unrelated_block_validator.py
```diff
@@ -0,0 +1,23 @@
...
+
+            if sorted(actions) != expected_actions:
```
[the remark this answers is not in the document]
ajmaddaford: Yes, that is how the unrelated block works.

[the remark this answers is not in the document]
MebinAbraham: Ya I just mean from runner's point, it doesn't care as long as the action exists. I.e if you had multiple options for yes/no like you have with driving collectors. Obv this requirement shouldn't chnage so not an issue.

app/validators/blocks/unrelated_block_validator.py
ajmaddaford: 👍

MebinAbraham (approved):

app/validators/blocks/unrelated_block_validator.py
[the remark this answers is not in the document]
ajmaddaford: I think validation gets unnecessarily more complicated, to allow for scenarios we have no use cases for.

PR #83, opened 2020-12-04
Add type check for relationships id

LJBabbage:
### PR Context
We currently allow the use of the block id **relationships** for non RelationshipCollectors, this will cause an error as it will hit the relationships end point and fail. This PR puts a check in to make sure the right type is being used
 

petechd (approved):

app/validators/blocks/block_validator.py
```diff
@@ -9,6 +9,7 @@ class BlockValidator(Validator):
     ANSWER_SELF_REFERENCE = "Invalid answer reference (self-reference)"
     COMPOSITE_ANSWER_INVALID = "Invalid composite answer"
     COMPOSITE_ANSWER_FIELD_INVALID = "Invalid field for composite answer"
+    ID_RELATIONSHIPS_NOT_USED_WITH_RELATIONSHIP_COLLECTOR = "Invalid use of id relationships, can only be used with type RelationshipCollector"
```
ajmaddaford, line 12:
```suggestion
    ID_RELATIONSHIPS_NOT_USED_WITH_RELATIONSHIP_COLLECTOR = "Invalid use of id relationships, can only be used with RelationshipCollector block type"
```

tests/schemas/invalid/test_invalid_use_of_id_relationships_with_type.json
```diff
@@ -0,0 +1,66 @@
+{
+  "mime_type": "application/json/ons/eq",
+  "language": "en",
+  "schema_version": "0.0.1",
+  "data_version": "0.0.3",
+  "survey_id": "0",
+  "title": "Test invalid use of id relationships",
+  "theme": "default",
+  "description": "A questionnaire to test invalid use of id relationships with type",
```
ajmaddaford, line 9: I'm not sure it's clear here what "with type" means, we could instead use:
> A questionnaire to test invalid use of block id relationships

The filename should also be updated if this is changed.

ajmaddaford: I think the filename should change to `test_invalid_use_of_block_id_relationships.json`

ajmaddaford (approved):

PR #86, opened 2021-01-06
Add boolean option to allow multiple lookups

petechd:
### PR Context
This adds new boolean option to question block that allows multiple lookups selection in textfield autosuggest component. It is required by [runner PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/487). This change is described on [Trello card](https://trello.com/c/RqJlAYzx). New invalid test schema was added.

### Checklist


petechd:
[FEATURE: Socio Demographic look-ups - Allow multiple selections (S)](https://trello.com/c/RqJlAYzx/4407-feature-socio-demographic-look-ups-allow-multiple-selections-s)

schemas/answers/text_field.json
```diff
@@ -22,6 +22,9 @@
       "suggestions_url": {
         "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
       },
+      "allow_multiple": {
+        "type": "boolean"
+      },
```
MebinAbraham, line 27: Could `allow_multiple` be conditional on the existence of `suggestions_url`?

MebinAbraham (approved):

schemas/answers/text_field.json
ajmaddaford: We should consider nesting `url` and `allow_multiple` under a `suggestions` key, then `url` can be mandatory if `suggestions` exist:

```json
"suggestions": {
  "url": "",
  "allow_multiple": ""
}
```

MebinAbraham: It would make the schema less ambiguous, so I think it would be beneficial to update.

petechd: I've added nested url in suggestions and dependency of suggestion on url, need to fix runner as schemas and resolve script are already fixed and seems like working with new validator.

schemas/answers/text_field.json
```diff
@@ -19,8 +19,21 @@
...
+        "properties": {
+          "url": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
+          },
+          "allow_multiple": {
+            "type": "boolean"
+          }
+        },
+        "dependencies": {
+          "suggestions": {
```
ajmaddaford, line 33: I don't think dependencies is needed as `url` is nested inside `suggestions`. Just marking `url` as required should only make it mandatory if the optional `suggestions` property exists.

petechd: I marked `url` as required now, dependency removed. Thanks!

ajmaddaford (approved):

tests/schemas/invalid/test_invalid_text_field_suggestions.json
```diff
@@ -0,0 +1,59 @@
+{
```
MebinAbraham, line 1: Is this schema needed?. We don't normally commit schemas to test raw JSON schema validation. As you are just testing the "required" rule I don't think we need this.

petechd: Schema removed, thanks!

MebinAbraham (approved):

MebinAbraham:
Don't merge this until runner and schemas are ready.

PR #88, opened 2021-04-07
Update python and packages

JamesGardiner:


petechd (approved):
Api and scripts working as expected👍

Pipfile
```diff
@@ -5,7 +5,7 @@ name = "pypi"
 
 [packages]
 Flask = "*"
-attrs = "==19.1.0"
+attrs = "==19.2.0"
 jsonpointer = "*"
 eq-translations = {editable = true,git = "https://github.com/ONSDigital/eq-translations.git",ref = "v2.0.0"}
```
MebinAbraham, line 10: Can't quite remember why we never upgraded this?

MebinAbraham (dismissed):

Pipfile
JamesGardiner: If we don't upgrade it, linting fails with an invalid kwarg argument error for the equality `eq` kwarg

MebinAbraham: I meant the eQ translations package.

ajmaddaford: We should remove this and see if it works as it's not a direct requirement. The reason for it being there was probably to fix a transitive dependency issue that has been resolved.

JamesGardiner: I'll give it a try

MebinAbraham (approved):

PR #90, opened 2021-04-16
Add remove final summary and confirmation blocks proposal

ajmaddaford:
### PR Context

Proposal to remove the `Summary` and `Confirmation` block types.

### Checklist

* [-] eq-translations updated to support any new schema keys which need translation

### DIFF
--- a/doc/decisions/0008-remove-final-summary-confirmation-blocks.md
+++ b/doc/decisions/0008-remove-final-summary-confirmation-blocks.md
@@ -21,6 +21,7 @@ The `SectionSummary` block type was [removed](https://github.com/ONSdigital/eq-q
   - an `options` object to hold options related to the given questionnaire flow:
     - For `Linear` questionnaires, an `include_summary` boolean option
     - For `Hub` questionnaires, the existing `required_completed_sections` option
+- Remove the top-level `hub` property
 
 ### Examples:
 
--- a/doc/decisions/0008-remove-final-summary-confirmation-blocks.md
+++ b/doc/decisions/0008-remove-final-summary-confirmation-blocks.md
@@ -19,7 +19,7 @@ The `SectionSummary` block type was [removed](https://github.com/ONSdigital/eq-q
 - Introduce a new `questionnaire_flow` top-level schema object with:
   - a `type` property to differentiate between `Linear` and `Hub` questionnaires
   - an `options` object to hold options related to the given questionnaire flow:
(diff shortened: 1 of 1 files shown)


doc/decisions/0008-remove-final-summary-confirmation-blocks.md
```diff
@@ -0,0 +1,66 @@
...
+
+The `SectionSummary` block type was [removed](https://github.com/ONSdigital/eq-questionnaire-validator/pull/29) for similar reasons. 
+
+## Proposal
+
+- Remove the `Summary` and `Confirmation` block types
+- Introduce a new `questionnaire_flow` top-level schema object with:
+  - a `type` property to differentiate between `Linear` and `Hub` questionnaires
+  - an `options` object to hold options related to the given questionnaire flow:
+    - For `Linear` questionnaires, an `include_summary_on_submission` boolean option
```
MebinAbraham, line 22: This should just be `inculde_summary`?. There is no other place this applies to and to me, it reads like it is on the submission click i.e. after submission rather than on final page.

MebinAbraham:
Regarding the URL routes, I am assuming that we will reuse the same `/questionnaire/` endpoint for the `Linear` flow, so that endpoint always displays the submission screen.

doc/decisions/0008-remove-final-summary-confirmation-blocks.md
JamesGardiner: I agree with @MebinAbraham - the `include_summary_on_submission` does lean towards suggesting it will be on submission click. I think `include_summary`, given the other constraints set out here, is enough.

ajmaddaford: I did have that originally. I'll change it back 👍

ajmaddaford:
> Regarding the URL routes, I am assuming that we will reuse the same `/questionnaire/` endpoint for the `Linear` flow, so that endpoint always displays the submission screen.

We should have a discussion about this before implementation in runner. I think we should consider having a submit endpoint for both hub and linear questionnaires. Currently for hub questionnaires POST requests to `/questionnaire` can either be to continue or submit; I think the code could be clearer with an explicit endpoint, and it would be more identifiable in logs.

MebinAbraham (approved):

JamesGardiner (approved):

PR #95, opened 2021-05-17
Remove Summary and Confirmation blocks + Introduce questionnaire flow

MebinAbraham:
### PR Context
Removed Summary and Confirmation blocks and introduced the questionnaire flow top level property.
Initial design doc: https://github.com/ONSdigital/eq-questionnaire-validator/pull/90

#### Decisions:

- __**Routing to the end of a Questionnaire**__
	- No new routing support has been added to route to the end of a questionnaire. Routing to the end can be achieved by:
		- Using 1 section and routing to the section "End".
		- Using multiple sections where a section routes to the section "End" and all other sections have the existing `enabled` rule.
	- There are certain caveats with routing to the end and using features like section summaries, at the moment, no additional validation has been added in this regard until business requirements are cleared up. Right now, the behavior of section summary with a final summary within a Linear survey is unchanged.
- Mandatory properties; Property such as `options` of questionnaire flow has been made mandatory to make schemas more explicit. However, if additional properties are specified then it is validated against the questionnaire flow schemas.

### How to review

- Do the changes make sense?
- Any required validation missing?

### Checklist


MebinAbraham:
I have pushed up [`8e372a7` (#95)](https://github.com/ONSdigital/eq-questionnaire-validator/pull/95/commits/8e372a7e6d89da9604d9af207791a1dbbf812c40) which conditionally includes the questionnaire flow options based on the type. The original commit [`770e36b` (#95)](https://github.com/ONSdigital/eq-questionnaire-validator/pull/95/commits/770e36b1d43425ba4d524eed1fd1a6135bc0cacf) is how is currently do it.

Take a look at the two and let me know your preference.

MebinAbraham:
Summary options are now nested to make the schema clearer. Options are now:
```json
       "options": {
            "summary": {
                "collapsible": false
            }
        }
```
`summary` property is optional, but `collapsible` is mandatory when summary is defined.

schemas/questionnaire_flow/linear.json
```diff
@@ -4,26 +4,19 @@
...
-      "include_summary": {
-        "description": "Whether to include the summary of answers",
-        "type": "boolean"
-      },
-      "collapsible": {
-        "description": "When True, the summary page will contain a list of collapsible section summaries",
-        "type": "boolean"
+      "summary": {
+        "type": "object",
+        "description": "Options for the summary",
```
ajmaddaford, line 9: I think the description should make it clear what the `summary` property is, something like:
```
Using this property will include a summary on the submit page.
```

schemas/questionnaire_flow/linear.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$id": "https://eq.ons.gov.uk/questionnaire_flow/linear.json",
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "options": {
+    "type": "object",
+    "properties": {
+      "summary": {
+        "type": "object",
+        "description": "Using this property will include a summary on the submit page",
```
LJBabbage, line 9: The description doesn't really match the pattern currently used, they are more statements, rather than directions. "Summary on submit page" is probably fine. Arguably you might not even need a description

LJBabbage (approved):
LGTM

ajmaddaford (approved):

PR #96, opened 2021-06-04
Add new schema for routing rules

ajmaddaford:
## Context

Adds schemas for the new rules documented in https://github.com/ONSdigital/eq-questionnaire-validator/blob/master/doc/decisions/0004-extend-and-improve-routing-rules.md, and makes them optional for `routing_rules`. This will allow us to incrementally implement the new rules in runner before deprecating the old format. 

Note that the usage of rules in skip conditions, section enabling and question/content variants has not been implemented in this pull request.

For now the only validator code that has been changed has been to:
- stop it from validating the new rules format
- allow self references in non placeholder value sources

## How to review

- Are the schemas structured appropriately?
- Are there any gaps?


tests/test_rules_schema.py
```diff
@@ -0,0 +1,25 @@
...
+from app.validators.schema_validator import SchemaValidator
+
+logger = getLogger()
+
+configure(logger_factory=LoggerFactory())
+
+
+def test_param_valid_schemas(rule_schema_filename):
+    """
+    Uses py.test generated tests to validate all schemas contained in the 'valid' folder.
```
MebinAbraham, line 16:
```suggestion
    Uses py.test generated tests to validate all schemas contained in the 'rules' folder.
```

schemas/rules/definitions.json
```diff
@@ -0,0 +1,93 @@
...
+      { "$ref": "https://eq.ons.gov.uk/rules/operators/date-offset.json" },
+      { "$ref": "https://eq.ons.gov.uk/rules/operators/boolean/and.json" },
+      { "$ref": "https://eq.ons.gov.uk/rules/operators/boolean/or.json" },
+      { "$ref": "https://eq.ons.gov.uk/rules/operators/boolean/not.json" }
+    ]
+  },
+  "any_value": {
+    "description": "Any value",
+    "oneOf": [
+      { "type": ["number", "string", "array", "null"] },
```
MebinAbraham, line 35: Shouldn't our `non-empty` string be used here?

schemas/rules/operators/comparison/all.json
```diff
@@ -0,0 +1,18 @@
...
+    "all": {
+      "type": "array",
+      "minItems": 2,
+      "maxItems": 2,
+      "items": {
+        "$ref": "https://eq.ons.gov.uk/rules/definitions.json#/array_value"
+      }
+    }
+  },
+  "required": ["all"],
```
MebinAbraham, line 16: Initially thought `all` and `any` read fine but since most of the operators are very flexible and because our `all` and `any` only works to check `in`, I think we should make them explicit, so `all-in` and `any-in` or even our current naming of `contains-all` and `contains-any` otherwise I think it gets confusing as you can't do `all([True, True, True])`. I think of `all` and `any` like Python's all and any and JSON schemas `allOf` and `anyOf` which our one does not resemble. The new `and` and `or` structure would match `all` and `any` exactly from a technical view. Thoughts?

MebinAbraham (commented):
Nice work

tests/conftest.py
```diff
@@ -14,3 +17,12 @@ def all_valid_schema_files():
...
     return schema_files
+
+
+def all_rule_schema_files():
+    schema_files = []
+    for folder, _, files in os.walk("tests/schemas/rules"):
+        for filename in files:
+            if filename.endswith(".json"):
+                schema_files.append(os.path.join(folder, filename))
+    return schema_files
```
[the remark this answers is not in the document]
ajmaddaford: This was copied from the block above, so I will update both.

schemas/rules/definitions.json
ajmaddaford: I don't think so, the rules will work with empty strings.

schemas/rules/operators/comparison/all.json
ajmaddaford: I think using `contains` would be confusing, as it's the opposite of `in` - `[a,b] contains b` vs `b in [a,b]`, so my preference would be to go for `all-in` and `any-in`. 

> The new and and or structure would match all and any exactly from a technical view

I don't know what you mean by this.

schemas/common_definitions.json
```diff
@@ -232,47 +232,107 @@
...
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+        "pattern": "^(\\d{4}-\\d{2}-\\d{2}|\\d{4}-\\d{2}|\\d{4}|now)$"
+      },
+      "source": {
+        "type": "string",
+        "enum": ["date"]
+      }
+    },
+    "required": ["value", "source"]
+  },
```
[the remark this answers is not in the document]
ajmaddaford: Done in https://github.com/ONSdigital/eq-questionnaire-validator/pull/96/commits/a79acc7e78a6ccd50e0198b0165ecccefd3e038c

schemas/rules/operators/comparison/any.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://eq.ons.gov.uk/rules/operators/comparison/any.json",
+  "description": "Evaluates if any values are in an array, takes two arguments",
```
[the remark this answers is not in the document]
ajmaddaford: Done in https://github.com/ONSdigital/eq-questionnaire-validator/pull/96/commits/ed3dd34371eed8e89070f71307bac1119171b11e

schemas/rules/definitions.json
```diff
@@ -0,0 +1,93 @@
...
+      },
+      {
+        "$ref": "https://eq.ons.gov.uk/rules/operators/comparison/less-than-or-equal.json"
+      },
+      {
+        "$ref": "https://eq.ons.gov.uk/rules/operators/comparison/more-than.json"
+      },
+      {
+        "$ref": "https://eq.ons.gov.uk/rules/operators/comparison/more-than-or-equal.json"
+      },
```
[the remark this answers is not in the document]
ajmaddaford: Done in https://github.com/ONSdigital/eq-questionnaire-validator/pull/96/commits/fa5758dc9805edfd1bfaee7aa0675177c8cae9ec

tests/conftest.py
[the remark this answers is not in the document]
ajmaddaford: This also highlighted a refactoring opportunity to use a shared method for finding json files 🙂 

Done in https://github.com/ONSdigital/eq-questionnaire-validator/pull/96/commits/fa5758dc9805edfd1bfaee7aa0675177c8cae9ec

tests/schemas/rules/boolean/and.json
```diff
@@ -0,0 +1,27 @@
...
+          "==": [
+            {
+              "source": "answers",
+              "identifier": "answer-2"
+            },
+            "Yes"
+          ]
+        }
+      ]
+    }
```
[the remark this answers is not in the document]
ajmaddaford: I have added test schema for the following:
- `answer-1 == 'Yes' and (answer-2 == 'Yes' or answer-3 == 'Yes')`
- `answer-1 == 'Yes' and (answer-2 not in ['No', 'Maybe'] or answer-3 == 'Yes')`

https://github.com/ONSdigital/eq-questionnaire-validator/pull/96/commits/4a97c52a13bb8dc89fcc26685bf2383f9a503d77

schemas/rules/operators/comparison/all.json
MebinAbraham: Ya fair point, therefore I'm happy to go with `all-in` and `any-in`.

> The new `and` and `or` structure would match all and any exactly from a technical view.

My bad, didn't word that right at all. All I meant by this is that our `and` and `or` operators pretty much does exactly what the proper `all()` and `any()` would do. i.e our JSON `and` is basically `and([True, True, False])` which is pretty much how `all` would work. So was just highlighting even more reason to be explicit by using `all-in` as we support `all` by supporting `and`. The same applies to `or` and `any`.

schemas/rules/definitions.json
MebinAbraham: I was wondering when that would be valid?. For unanswered it would be `null`, so we wouldn't ever compare empty strings?

ajmaddaford: My thinking is that is not something that would be known until runtime of an implementation, and the rules are not tied to our implementation. I also think it makes sense to make it as permissive as possible for ease of extending it in the future.

schemas/rules/operators/comparison/all.json
ajmaddaford: Now I understand 🙂 

I will update to use `all-in` and `any-in`.

schemas/rules/operators/comparison/less-than.json
```diff
@@ -0,0 +1,18 @@
...
+  "$id": "https://eq.ons.gov.uk/rules/operators/comparison/less-than.json",
+  "description": "Evaluates less than, takes two arguments",
+  "type": "object",
+  "properties": {
+    "<": {
+      "type": "array",
+      "minItems": 2,
+      "maxItems": 2,
+      "items": {
+        "$ref": "https://eq.ons.gov.uk/rules/definitions.json#/any_value_or_operator"
```
MebinAbraham, line 12: As discussed, we should look at restricting these to any value or non-boolean operators instead of all operators.

schemas/rules/definitions.json
```diff
@@ -0,0 +1,97 @@
...
+    ]
+  },
+  "array_or_string_value": {
+    "description": "Any array or string value",
+    "oneOf": [
+      { "type": ["string", "array"] },
+      {
+        "$ref": "https://eq.ons.gov.uk/value_sources.json#/answer_value_source"
+      }
+    ]
```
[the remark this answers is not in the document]
ajmaddaford: I disagree as I think it makes it less flexible for future changes. If we had a `string_value` type that is a combination of `string` with `answer_value_source`, then `array_or_string_value` couldn't use both `array_value` and `string_value` as both types contain `answer_value_source`.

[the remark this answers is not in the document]
MebinAbraham: Right, ya, that makes sense. 👍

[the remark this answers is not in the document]
ajmaddaford: Ignore that, I assumed `anyOf` didn't work when there is an overlap in the refs. I am now exploring removing the `array_or_string_value` and just using `array_value` and `string_value`.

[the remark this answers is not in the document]
ajmaddaford: This is now redundant due to changes related to your other comment below.

schemas/rules/operators/comparison/less-than.json
ajmaddaford: Done in https://github.com/ONSdigital/eq-questionnaire-validator/pull/96/commits/8da0b1f405624c0861dd104460b9043e11ec9332

schemas/rules/operators/comparison/in.json
```diff
@@ -0,0 +1,23 @@
...
+      "maxItems": 2,
+      "items": [
+        {
+          "$ref": "https://eq.ons.gov.uk/rules/definitions.json#/string_value"
+        },
+        {
+          "$ref": "https://eq.ons.gov.uk/rules/definitions.json#/array_value"
+        }
+      ]
+    }
```
MebinAbraham, line 19: Shouldn't the first value be `any_value` as it could be other types like a number and also the current definition wouldn't support values from other sources like metadata. I know `any_value` could be an array, but I still think that is fine to accept for this case.

schemas/rules/operators/comparison/greater-than-or-equal.json
```diff
@@ -0,0 +1,28 @@
...
+      "minItems": 2,
+      "maxItems": 2,
+      "items": {
+        "anyOf": [
+          {
+            "$ref": "https://eq.ons.gov.uk/rules/definitions.json#/number_value"
+          },
+          {
+            "$ref": "https://eq.ons.gov.uk/rules/definitions.json#/date_value"
+          },
```
MebinAbraham, line 18: We should support `any_value` here as we would want this usable with metadata source etc?. Applies to other schemas where this is used also. I think there is only so much we can do in schemas, but we should allow the likes of metadata as a source, even if that could be of any type. I think at schema level as long as we validate it is one of the value sources we support, then it's fine.

app/validators/blocks/block_validator.py
```diff
@@ -45,23 +46,24 @@ def validate_id_relationships_used_with_relationship_collector(self):
...
+        if isinstance(source_reference["identifier"], str):
+            identifiers = [source_reference["identifier"]]
+        else:
+            identifiers = source_reference["identifier"]
+
+        if source == "answers":
+            selector = source_reference.get("selector")
+            self.validate_answer_source_reference(
+                source_reference_path, identifiers, selector
+            )
```
LJBabbage, line 60: I find this hard to follow with all the sources in the vars and then source_reference having a source, can we possibly make them more distinguishable? Due to the naming my mind also tells me sources_reference is an item of the list sources references, which it isn't in this instance.

schemas/rules/operators/boolean/and.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://eq.ons.gov.uk/rules/operators/boolean/and.json",
+  "description": "Evaluates boolean `and`, takes two or more arguments",
```
LJBabbage, line 4: Would a full stop not make this read better rather than a comma? Does it need the boolean in the description, considering all these conditions are in the boolean folder? Maybe "Evaluates `and` operator. Takes two or more arguments

schemas/rules/operators/boolean/not.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://eq.ons.gov.uk/rules/operators/boolean/not.json",
+  "description": "Evaluates boolean `not`, takes one argument",
```
LJBabbage, line 4: As above, I won't mark any others

LJBabbage (approved):
Awesome job, all I can find is a few grammatical suggestions and a small refactor, which are of minimal importance and not worth blocking over. A few descriptions are a bit short and sweet like "Two values" which wasn't immediately obvious what it meant. 

I couldn't see any obvious gaps, but feel once we start migrating schemas in runner it will be easier to spot.

app/validators/blocks/block_validator.py
ajmaddaford: I agree, but the only thing changed here was move the loop up a level, and this is likely to be changed in the validator code changes I'm making.

schemas/rules/operators/comparison/in.json
ajmaddaford: To address this I have created a `any_value_except_array` in https://github.com/ONSdigital/eq-questionnaire-validator/pull/96/commits/2c073ca22ce6e90ac03b8b4e943c6bf121e3a048.

schemas/rules/operators/comparison/greater-than-or-equal.json
ajmaddaford: As these operators won't support arrays, I have used the `any_value_except_array` in https://github.com/ONSdigital/eq-questionnaire-validator/pull/96/commits/2c073ca22ce6e90ac03b8b4e943c6bf121e3a048

schemas/rules/operators/value/date.json
```diff
@@ -0,0 +1,55 @@
...
+        {
+          "oneOf": [
+            {
+              "description": "A date value in the format YYYY-MM-DD or `now`",
+              "type": "string",
+              "pattern": "^(\\d{4}-\\d{2}-\\d{2}|\\d{4}-\\d{2}|\\d{4}|now)$"
+            },
+            {
+              "$ref": "https://eq.ons.gov.uk/value_sources.json#/any_value_source"
+            }
```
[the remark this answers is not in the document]
ajmaddaford: No, the type above allows date strings and `now`, this allows dates from metadata and answers.

MebinAbraham (approved):
👍

schemas/rules/operators/value/date.json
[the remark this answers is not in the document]
MebinAbraham: Whoops, misread it.

schemas/rules/operators/boolean/and.json
ajmaddaford: I have removed the bit about number of arguments from all schemas as it is already documented by the schema restrictions - https://github.com/ONSdigital/eq-questionnaire-validator/pull/96/commits/9802ad6280387a8199a7233aa661ca16f8c48ad2

schemas/rules/operators/boolean/not.json
ajmaddaford: As above.

ajmaddaford:
@LJBabbage Since you have reviewed I have made some changes to the way dates work in https://github.com/ONSdigital/eq-questionnaire-validator/pull/96/commits/2c073ca22ce6e90ac03b8b4e943c6bf121e3a048. Dates are now handled by a `date` operator rather than as a value source. This was done for two reasons:
- we need to know when to treat a value as a date that isn't obviously a date e.g. a metadata date
- the value source implementation was inconsistent with the other value sources (`source` and `identifier` vs `source` and `value`)

schemas/rules/operators/value/date.json
```diff
@@ -1,16 +1,25 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema",
-  "$id": "https://eq.ons.gov.uk/rules/operators/value/date-offset.json",
-  "description": "Offsets a date, takes two arguments",
+  "$id": "https://eq.ons.gov.uk/rules/operators/value/date.json",
+  "description": "Casts a value to a date, first argument is a date, second is an offset",
```
LJBabbage: Would this not pass if I provided just an offset? Also do we ever say arguments are optional in the description?

LJBabbage (changes requested):
Just one query

LJBabbage (approved):

schemas/rules/operators/value/date.json
ajmaddaford: As discussed the date is mandatory as it's the first item in the array. I have updated the text to make it explicit that the offset is optional.

PR #97, opened 2021-06-24
Add code validation of new routing rules

ajmaddaford:
### PR Context

Adds code validation of the new routing rules to cover what the schema doesn't validate. The existing routing rules validation remains, and will be removed when we move fully to the new routing rules format.

Main changes:
- `AnswerRoutingValidator` has been removed. This didn't do anything as it only validates routing rules with no defaults, which can't happen as we enforce a default rule.
- The `NewRoutingValidator` is very similar to the existing `RoutingValidator`, other than some refactoring the only change is the removal of the unnecessary `goto` schema property.
- Value source reference validation has been moved from the `BlockValidator` into it's own `ValueSourceValidator`. This is now called at the section level, so that it also validates references in group routing rules.
- Most of the new rule validation is done in the `NewWhenRuleValidator`. 
  - In order to validate the rules, the notion of types and resolution of operator arguments to types has been introduced. 
  - For any given operation we resolve the types involved, and validate that the types are valid for the given operation. 
  - The type resolution resolves two main things, Python types (fixed value(s) in the schema) and value source references (answers, metadata, lists etc). 
  - The types used are based on JSON types. This was chosen because:
    - we need a common type to be able to compare the types of value source references with the types of fixed values
    - answers in runner are stored as JSON
    - the fixed values in the schema are JSON types
    - there is a common grouping for `number` (rather than `int`, `float` etc in Python)

### Checklist


app/validators/routing/new_routing_validator.py
```diff
@@ -0,0 +1,77 @@
+from app.validators.routing.new_when_rule_validator import NewWhenRuleValidator
+from app.validators.validator import Validator
+
+
+class NewRoutingValidator(Validator):
+    ROUTING_RULES_DO_NOT_HAVE_A_DEFAULT_RULE = (
+        "Routing rules do not have a default rule"
+    )
+    ROUTING_RULES_HAS_TOO_MANY_DEFAULTS = "Routing rules has more than one default rule"
```
berroar, line 9:
```suggestion
    ROUTING_RULES_HAS_TOO_MANY_DEFAULTS = "Routing rules have more than one default rule"
```

app/validators/routing/new_routing_validator.py
```diff
@@ -0,0 +1,77 @@
...
+                )
+            self.validate_routing_rule(rule)
+
+        return self.errors
+
+    def validate_routing_rules_have_default(self, rules):
+        """
+        Ensure that a set of routing rules contains a default, without a when clause.
+        """
+        default_routing_rule_count = 0
```
berroar, line 40: Could use list comprehension here?
```suggestion
        default_routing_rule_count = sum("when" not in rule for rule in rules)
```

ajmaddaford: Neat! I didn't realise `sum` could be used to count booleans.

app/validators/routing/new_when_rule_validator.py
```diff
@@ -0,0 +1,196 @@
...
+COMPARISON_OPERATORS = [
+    OPERATOR_EQUAL,
+    OPERATOR_NOT_EQUAL,
+    OPERATOR_GREATER_THAN,
+    OPERATOR_LESS_THAN,
+    OPERATOR_GREATER_THAN_OR_EQUAL,
+    OPERATOR_LESS_THAN_OR_EQUAL,
+]
+
+ARRAY_OPERATORS = [OPERATOR_IN, OPERATOR_ALL_IN, OPERATOR_ANY_IN]
```
LJBabbage, line 38: I wonder if this this should have 'in' in the var somewhere, down if the _if conditions_ I had to come back up to check what the array operators included to immediate think all the _ins_

app/validators/routing/new_when_rule_validator.py
```diff
@@ -0,0 +1,196 @@
...
+OPERATOR_GREATER_THAN = ">"
+OPERATOR_LESS_THAN = "<"
+OPERATOR_GREATER_THAN_OR_EQUAL = ">="
+OPERATOR_LESS_THAN_OR_EQUAL = "<="
+OPERATOR_IN = "in"
+OPERATOR_ALL_IN = "all-in"
+OPERATOR_ANY_IN = "any-in"
+OPERATOR_DATE = "date"
+
+BOOLEAN_OPERATORS = [OPERATOR_NOT, OPERATOR_AND, OPERATOR_OR]
```
LJBabbage, line 27: Boolean always makes me think true false, is logic operators not more appropriate?

app/validators/routing/new_routing_validator.py
ajmaddaford: Done 👍

app/validators/routing/new_when_rule_validator.py
```diff
@@ -0,0 +1,196 @@
...
+            and TYPE_NULL not in argument_types
+        ):
+            self._validate_argument_types_match(rule, argument_types)
+
+        if operator_name in ARRAY_OPERATORS + [OPERATOR_EQUAL, OPERATOR_NOT_EQUAL]:
+            self._validate_options(rule, operator_name)
+
+        if operator_name == OPERATOR_DATE:
+            return TYPE_DATE
+        return TYPE_BOOLEAN
```
LJBabbage, line 104: This is fun, like a mathematic guess who, but although a bit hard to digest does the job. Can't say I have gone through all combinations mind, which is a slight concern, might loop back later.

LJBabbage (commented):
On first pass this looks good, only arbitrary comments. The extra refactoring around source and metadata is an improvement but unless I am missing something is not directly related to this story, perfectly fine to leave mind.

app/validators/sections/section_validator.py
```diff
@@ -40,23 +43,65 @@ def validate_list_exists(self, list_name):
...
+        for json_path, source_reference in source_references:
+            if "source" in source_reference:
+                value_source_validator = ValueSourceValidator(
+                    value_source=source_reference,
+                    json_path=json_path,
+                    questionnaire_schema=self.questionnaire_schema,
+                )
+                self.errors += value_source_validator.validate()
+
     def validate_groups(self):
```
petechd, line 64: This could be refactored as it shares the same logic as the code underneath, we can add a new method like this:
```suggestion
    def validate_groups(self):
        for group in self.section["groups"]:
            self.validate_unit(group, group)
            self.validate_blocks(group["id"])

    def validate_blocks(self, group_id):
        group = self.questionnaire_schema.get_group(group_id)

        for block in group.get("blocks"):
            self.validate_unit(block, group)

            block_validator = get_block_validator(block, self.questionnaire_schema)
            self.errors += block_validator.validate()

            self.validate_question(block)
            self.validate_variants(block)

    def validate_unit(self, unit, group):
        if "routing_rules" in unit:
            if any("goto" in rule for rule in unit["routing_rules"]):
                block_routing_validator = RoutingValidator(
                    unit, group, self.questionnaire_schema
                )
            else:
                block_routing_validator = NewRoutingValidator(
                    routing_rules=unit["routing_rules"],
                    group=group,
                    origin_id=unit["id"],
                    questionnaire_schema=self.questionnaire_schema,
                )
            self.errors += block_routing_validator.validate()

        if "skip_conditions" in unit:
            self.validate_skip_conditions(unit["skip_conditions"], unit["id"])

```

app/validators/routing/new_routing_validator.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def __init__(self, routing_rules, group, origin_id, questionnaire_schema):
+        super().__init__(routing_rules)
+        self.routing_rules = routing_rules
+        self.group = group
+        self.origin_id = origin_id
+        self.questionnaire_schema = questionnaire_schema
+
+    def validate(self):
+        self.validate_routing_rules_have_default(self.routing_rules)
```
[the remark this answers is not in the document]
ajmaddaford: I have renamed the method `validate_routing_rules_has_single_default_rule` and updated the docstring to be clearer in https://github.com/ONSdigital/eq-questionnaire-validator/pull/97/commits/902935236819eea573fe443627d3623da7bccb28

app/validators/routing/new_routing_validator.py
```diff
@@ -0,0 +1,77 @@
...
+        default_routing_rule_count = 0
+        for rule in rules:
+            if "when" not in rule:
+                default_routing_rule_count += 1
+
+        if not default_routing_rule_count:
+            self.add_error(
+                self.ROUTING_RULES_DO_NOT_HAVE_A_DEFAULT_RULE, origin_id=self.origin_id
+            )
+        elif default_routing_rule_count > 1:
```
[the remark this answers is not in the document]
ajmaddaford: `default_routing_rule_count` was changed to use `sum` and list comprehension based on comments from @berroar. Although slightly less efficient, I think it's clearer.

app/validators/routing/new_when_rule_validator.py
```diff
@@ -0,0 +1,196 @@
...
+            if isinstance(argument, dict) and any(
+                operator in argument for operator in ALL_OPERATORS
+            ):
+                argument_type = self.validate_rule(argument)
+            elif isinstance(argument, dict) and "source" in argument:
+                argument_type = resolve_value_source_json_type(
+                    argument, self.questionnaire_schema.answers_with_context
+                )
+            else:
+                argument_type = python_type_to_json_type(type(argument).__name__)
```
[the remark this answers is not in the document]
ajmaddaford: The second argument for the `date` operator is an object containing the offset values e.g.
```
{
  "years": 1,
  "months": 10
}
```

app/validators/routing/new_when_rule_validator.py
ajmaddaford: I named it this to make it clear that these are operators that work with arrays. I don't think we can combine `ARRAY` with `IN` in the name as the order implies something i.e. `ARRAY_IN` vs `IN_ARRAY`. We could name it `IN_OPERATORS`.

app/validators/routing/new_when_rule_validator.py
ajmaddaford: Happy to go with `LOGIC_OPERATORS`.

ajmaddaford: I have changed this in https://github.com/ONSdigital/eq-questionnaire-validator/pull/97/commits/e7d9c17c0627a151002290fcccd6bee2fde90398. As part of this I have also renamed `boolean` to `logic` in the schemas.

app/validators/routing/new_when_rule_validator.py
ajmaddaford: This isn't meant to exhaustively validate all the operations. There could be an operator that has no need for any code validation.

app/validators/sections/section_validator.py
ajmaddaford: Good shout. I have called the common method `validate_routing` and used `schema_element` in place of `unit` in https://github.com/ONSdigital/eq-questionnaire-validator/pull/97/commits/41eaf12b93a601c7ebb325e28d2627890aa53458

app/validators/routing/new_when_rule_validator.py
LJBabbage: IN_OPERATORS for me

app/validators/sections/section_validator.py
petechd: Thanks👍

app/validators/routing/new_when_rule_validator.py
ajmaddaford: I was going to make the change, but after trying it I'm not convinced. As with the rename of `boolean` to `logic`, if we rename this we should also rename `array` to `in` in the schemas; this leads to paths of `in/in.json`, which I think is less clear then `array/in.json`

LJBabbage (approved):

petechd (approved):

PR #100, opened 2021-07-08
Add design proposal for telephone and email placeholder transforms

ajmaddaford:
### PR Context

Design proposal for adding telephone and email placeholder transforms.

### Checklist

### DIFF
--- a/doc/decisions/0009-telephone-and-email-placeholder-transforms.md
+++ b/doc/decisions/0009-telephone-and-email-placeholder-transforms.md
@@ -1,0 +1,73 @@
+# 9. Telephone and email placeholder transforms
+
+## Context
+
+When rendering telephone numbers and email addresses, we need to be able to wrap them with specific markup. This markup should not be included in schema text as it:
+- binds the schema to the output format 
+- creates translation issues
+
+## Proposal
+
+Introduce two new placeholder transforms that abstract the markup from the text.  
+
+### Telephone number link transform
+
+The `telephone_number_link` transform will allow runner to wrap the number with a `tel` href.
+
+```json
+{
+  "text": "Ring us on {telephone_number}",
+  "placeholders": [
+    {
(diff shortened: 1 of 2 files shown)


doc/decisions/0009-telephone-and-email-placeholder-transforms.md
```diff
@@ -0,0 +1,73 @@
...
+  "text": "Email us at {email_address}",
+  "placeholders": [
+    {
+      "placeholder": "email_address",
+      "transforms": [
+        {
+          "transform": "email_link",
+          "arguments": {
+            "email_address": "surveys@ons.gov.uk",
+            "email_subject": "Change of details reference ",
```
MebinAbraham, line 53: Makes sense for the time being. Do we want to allow defining the text for the href also, or do we want to enforce it will be same as tel/email itself? Either way, translation would be an issue until we do like recursive placeholders.

LJBabbage (approved):
As in the meeting there seems to be many ways this could be achieved, I prefer the generic link route, but this is ok

doc/decisions/0009-telephone-and-email-placeholder-transforms.md
ajmaddaford: The current requirement is that the link text is always the phone number or email address, so didn't made it customisable. There is no translation issue with this as the phone number or email address don't need to be translated.

pricem14pc (approved):
Aligns with the tech session discussions as the more popular approach

doc/decisions/0009-telephone-and-email-placeholder-transforms.md
pricem14pc: I think the best practice is for the link text (telephone / email address) to be the value of the target so this is fine.

MebinAbraham (approved):

PR #102, opened 2021-07-14
Add email and telephone transforms

LJBabbage:
### PR Context
This PR adds telephone number and email link transforms

### Checklist


schemas/string_interpolation/transforms/email_link.json
```diff
@@ -0,0 +1,31 @@
...
+    "properties": {
+      "transform": {
+        "type": "string",
+        "enum": ["email_link"]
+      },
+      "arguments": {
+        "type": "object",
+        "properties": {
+          "email_address": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
```
ajmaddaford, line 15: Email address validation could be done in schema rather than code - https://json-schema.org/understanding-json-schema/reference/string.html#email-addresses

schemas/string_interpolation/transforms/telephone_number_link.json
```diff
@@ -0,0 +1,25 @@
...
+    "properties": {
+      "transform": {
+        "type": "string",
+        "enum": ["telephone_number_link"]
+      },
+      "arguments": {
+        "type": "object",
+        "properties": {
+          "telephone_number": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
```
ajmaddaford, line 15: The card says basic validation should be done on the phone number - a simple regex in the schema for numbers, `+` and spaces should be enough.

schemas/string_interpolation/transforms/email_link.json
petechd: Something like:
```
  "email": {
    "description": "",
    "pattern": "^[A-Z0-9._%+-]+[A-Z0-9.-]+\\.[A-Z]{2,}$"
  },
```

petechd: Similar validation for telephone number...

LJBabbage: True but the card said to use the runner implementation which is why I used email-validator

schemas/string_interpolation/transforms/email_link.json
```diff
@@ -0,0 +1,31 @@
...
+    "properties": {
+      "transform": {
+        "type": "string",
+        "enum": ["email_link"]
+      },
+      "arguments": {
+        "type": "object",
+        "properties": {
+          "email_address": {
+            "pattern": "^[A-Za-z0-9._+-]+@[A-Za-z0-9.-]+.[A-Za-z.]+$"
```
ajmaddaford, line 15: There should also be a `type` property set to `string`.
https://json-schema.org/understanding-json-schema/reference/string.html#regular-expressions

schemas/string_interpolation/transforms/telephone_number_link.json
```diff
@@ -0,0 +1,26 @@
...
+      "transform": {
+        "type": "string",
+        "enum": ["telephone_number_link"]
+      },
+      "arguments": {
+        "type": "object",
+        "properties": {
+          "telephone_number": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+            "pattern": "^[0-9 ]+$"
```
ajmaddaford, line 16: I don't think the ref can be combined with the regex. This still allows letters, changing `$ref` to `type: string` fixes the issue (and doesn't allow empty strings).

PR #103, opened 2021-07-22
Add view_submitted_response

LJBabbage:
### PR Context
Introduce view_submitted_response as an optional boolean

### Checklist


MebinAbraham (approved):

MebinAbraham:
Given it is already under the submission key, maybe it should be like `view_answers/view_response` instead of `view_submitted_response`. Keep it as is I guess, unless others have any other views.

berroar (approved):

PR #104, opened 2021-08-10
Validate answer label is present when using multiple answer

MebinAbraham:
### PR Context
- Adds validation to ensure that questions with multiple answers have an answer label except for `MutuallyExclusive`.
- Fixed some linting that was missed.

Ensure the changes make sense.

### Checklist

### DIFF
--- a/app/validators/questions/question_validator.py
+++ b/app/validators/questions/question_validator.py
@@ -23,15 +23,11 @@ class QuestionValidator(Validator):
         if len(self.answers) < 2:
             return None
 
+        final_answer_single_checkbox = self.answers[1]["type"] == "Checkbox"
         for answer in self.answers:
-            if not answer.get("label"):
-                if (
-                    answer["type"] == "Checkbox"
-                    and (answer == self.answers[-1])
-                    and len(answer["options"]) < 2
-                ):
-                    return None
-
+            if not (
+                len(self.answers) == 2 and final_answer_single_checkbox
+            ) and not answer.get("label"):
                 self.add_error(
                     self.ANSWER_LABEL_MISSING_MULTIPLE_ANSWERS, answer_id=answer["id"]
                 )
--- a/app/validators/questions/question_validator.py
+++ b/app/validators/questions/question_validator.py
@@ -25,6 +25,13 @@ class QuestionValidator(Validator):
 
         for answer in self.answers:
             if not answer.get("label"):
+                if (
+                    answer["type"] == "Checkbox"
+                    and (answer == self.answers[-1])
+                    and len(answer["options"]) < 2
+                ):
+                    return None
+
                 self.add_error(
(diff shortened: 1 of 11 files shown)


app/validators/questions/question_validator.py
```diff
@@ -2,9 +2,36 @@
...
+        if len(self.answers) < 2:
+            return None
+
+        for answer in self.answers:
+            if not answer.get("label"):
+                if (
+                    answer["type"] == "Checkbox"
+                    and (answer == self.answers[-1])
+                    and len(answer["options"]) < 2
+                ):
```
ajmaddaford, line 32: I think a single checkbox should only be used like this when there are two answers, although this doesn't seem to be documented at all in the design system.

This also enforces a label for the first answer, which works for the current schemas as the first answer is a `Number`. This pattern may need to be used with other answer types e.g. Radio, so I think we should allow the first answer to not have a label.

tests/validators/questions/test_question_validator.py
```diff
@@ -0,0 +1,109 @@
+from app.validators.questions import get_question_validator
+
+
+def test_missing_answer_label_single_answer_ignored():
```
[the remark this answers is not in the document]
MebinAbraham: Updated.

app/validators/questions/question_validator.py
```diff
@@ -23,15 +23,11 @@ def _validate_answer_labels(self):
...
-                if (
-                    answer["type"] == "Checkbox"
-                    and (answer == self.answers[-1])
-                    and len(answer["options"]) < 2
-                ):
-                    return None
-
+            if not (
+                len(self.answers) == 2 and final_answer_single_checkbox
+            ) and not answer.get("label"):
```
ajmaddaford, line 30: The checkbox is checked in every loop, even though it won't change. It should be done in a conditional outside of the loop. It should also check that it's a single checkbox:
```python
        if (
            len(self.answers) == 2
            and self.answers[-1]["type"] == "Checkbox"
            and len(self.answers[-1]["options"]) == 1
        ):
            return None
```

tests/validators/questions/test_question_validator.py
```diff
@@ -0,0 +1,112 @@
...
+        ],
+    }
+
+    validator = get_question_validator(question)
+    validator.validate()
+
+    assert not validator.errors
+
+
+def test_no_answer_label_max_2_answers_last_answer_checkbox_ignored():
```
ajmaddaford, line 83:
```suggestion
def test_no_answer_label_two_answers_last_answer_single_checkbox_ignored():
```

ajmaddaford (approved):

LJBabbage (approved):
Looks fine, probably wouldn't have used the word ignored as it didn't add anything, and I still had to look to see it just doesn't validate labels existence, however it's arbitrary. Only other things is a small typo M1000

PR #110, opened 2021-09-22
Routing on checkbox count

rmccar:
### PR Context
Adds the validation for the count operator

### Checklist
- Tests pass
- [New schema](https://github.com/ONSdigital/eq-questionnaire-runner/pull/670) in runner is validated


rmccar:
[Routing on count of checkbox answer (2 SP)](https://trello.com/c/q9ngIbPM/4796-routing-on-count-of-checkbox-answer-2-sp)

MebinAbraham (commented):
The same way the date operator checks that any referenced answers are dates types, we need to ensure any references to answers when using `count` is of type `Checkbox`. https://github.com/ONSdigital/eq-questionnaire-validator/blob/master/app/validators/routing/new_when_rule_validator.py#L136

schemas/rules/operators/value/count.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://eq.ons.gov.uk/rules/operators/value/count.json",
+  "description": "Returns the number of objects in an array. Accepts a single arguement which is the array to get the number of objects from",
```
LJBabbage, line 4: Minor spelling mistake with "arguement". Would probably be better using the term count rather than number to align with the operator name. I would shorten it slightly, but up to you
```
Returns the count of items in an array. Accepts a single argument which is the array
```

LJBabbage (changes requested):

LJBabbage (approved):

app/validators/routing/new_when_rule_validator.py
```diff
@@ -101,6 +105,9 @@ def validate_rule(self, rule):
 
         if operator_name == OPERATOR_DATE:
             return TYPE_DATE
+
+        if operator_name == OPERATOR_COUNT:
+            self._validate_count_operator(rule)
```
MebinAbraham, line 110: This needs to be done the same way as the date one. You need to first validate the operator then return the `type` for the operator. At the moment, this code would raise errors against your runner schema. This also highlight missing tests, we have tests to validate types for dates, so we need to do the same for count.

schemas/string_interpolation/transforms/list_item_count.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$id": "https://eq.ons.gov.uk/string_interpolation/transforms/list_item_count.json",
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "list_item_count": {
```
MebinAbraham, line 4: This transform should also be added to `test_string_transforms.json` schema.

app/validators/routing/new_when_rule_validator.py
```diff
@@ -101,8 +97,26 @@ def validate_rule(self, rule):
...
+                operator in argument for operator in ALL_OPERATORS
+            ):
+                argument_type = self.validate_rule(argument)
+            elif isinstance(argument, dict) and "source" in argument:
+                argument_type = resolve_value_source_json_type(
+                    argument, self.questionnaire_schema.answers_with_context
+                )
+            else:
+                argument_type = python_type_to_json_type(type(argument).__name__)
+            argument_types.append(argument_type)
```
MebinAbraham, line 118: The main job of this is to return the arg types.
- We should rename this to something like `_get_argument_types_for_operator`, also should be private.
- We shouldn't pass in `argument_types`, it should just return it the list of arg types, we shouldn't mutate what is being passed in. So I would change the signature `def _get_argument_types_for_operator(self, arguments):`

schemas/rules/operators/value/count.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://eq.ons.gov.uk/rules/operators/value/count.json",
+  "description": "Returns the count of items in an array. Accepts a single argument which is the array",
+  "type": "object",
+  "properties": {
+    "count": {
```
MebinAbraham, line 7: We should also add this to the test rules. `tests/schemas/rules/value/`.

app/validators/routing/new_when_rule_validator.py
```diff
@@ -68,23 +72,14 @@ def validate(self):
...
-            ):
-                argument_type = self.validate_rule(argument)
-            elif isinstance(argument, dict) and "source" in argument:
-                argument_type = resolve_value_source_json_type(
-                    argument, self.questionnaire_schema.answers_with_context
-                )
-            else:
-                argument_type = python_type_to_json_type(type(argument).__name__)
-            argument_types.append(argument_type)
+        argument_types = self._get_argument_types_for_operator(rule, operator_name)
```
MebinAbraham, line 75: Not a blocker, this could pass in the actual operand as `rule[operator_name]` since those are not needed indivudally.

MebinAbraham (approved):
👍

PR #113, opened 2021-09-29
Adds validators for generate_date_range and format_date_range_pair

jamesnunn:
### PR Context
Adds validators for generate_date_range and format_date_range_pair

### Checklist


jamesnunn:
[Using dynamic dates as placeholders (5 SP)](https://trello.com/c/e6WOAWXz/4866-using-dynamic-dates-as-placeholders-5-sp)

petechd (approved):

schemas/string_interpolation/transforms/generate_date_range.json
```diff
@@ -0,0 +1,74 @@
+{
+  "$id": "https://eq.ons.gov.uk/string_interpolation/transforms/generate_date_range.json",
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "generate_date_range": {
+    "description": "A transform for generating the start and end date of a date range from a reference date, offset number of weeks, number of days in range and an optional first day of the week.",
+    "type": "object",
+    "properties": {
+      "transform": {
+        "type": "string",
+        "enum": ["generate_date_range", "format_date_range_pair"]
```
petechd, line 10: Two transforms names in here, needs fixing.

petechd (changes requested):

petechd (approved):

schemas/string_interpolation/transforms/format_date_range_pair.json
```diff
@@ -0,0 +1,36 @@
+{
+  "$id": "https://eq.ons.gov.uk/string_interpolation/transforms/format_date_range_pair.json",
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "format_date_range": {
+    "description": "A transform for formatting a pair of dates representing the start and end of a date range.",
+    "type": "object",
+    "properties": {
+      "transform": {
+        "type": "string",
+        "enum": ["format_date_range_pair"]
```
MebinAbraham, line 10: The `_pair` could be dropped, I don't think it is necessary.

schemas/string_interpolation/transforms/generate_date_range.json
MebinAbraham, line 10: How come both transforms are defined here?.

schemas/string_interpolation/transforms/generate_date_range.json
```diff
@@ -0,0 +1,74 @@
+{
+  "$id": "https://eq.ons.gov.uk/string_interpolation/transforms/generate_date_range.json",
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "generate_date_range": {
```
MebinAbraham, line 4: Could just be called `date_range`, to me, `generate` doesn't add much plus proposal for date range in the new rules structure uses `date-range` which placeholders would eventually be refactored to use.

schemas/string_interpolation/transforms/generate_date_range.json
```diff
@@ -0,0 +1,74 @@
...
+              {
+                "$ref": "https://eq.ons.gov.uk/common_definitions.json#/date_value"
+              }
+            ]
+          },
+          "offset_full_weeks": {
+            "oneOf": [
+              {
+                "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/value_sources"
+              },
```
MebinAbraham, line 29: Adding value sources into these inputs extends the scope as it is now within the current requirements. Happy to keep as it gives us the flexibility, just need to ensure tests these in runner with value sources as well as literal ints etc.

However, we would need to add additional python validation, i.e when an `answer` source is used, we need to validate that only a `Number` type is used and that it has a min value of `1`. etc. Same applies for `days_in_range`.

schemas/string_interpolation/transforms/generate_date_range.json
```diff
@@ -0,0 +1,74 @@
...
+                "default": "monday",
+                "enum": [
+                  "monday",
+                  "tuesday",
+                  "wednesday",
+                  "thursday",
+                  "friday",
+                  "saturday",
+                  "sunday"
+                ]
```
MebinAbraham, line 62: In v2, we used abbreviation for days i.e `MO`, `TU`, in the proposal I did for v3 dynamic option, I used abbrev like `MON`, `TUE` etc. I think we should align. I'm not against using the full string, but think constants should be uppercase. Also, what is your preference, `MON` etc or `MONDAY`?

schemas/string_interpolation/transforms/generate_date_range.json
```diff
@@ -0,0 +1,74 @@
...
+        "enum": ["generate_date_range", "format_date_range_pair"]
+      },
+      "arguments": {
+        "type": "object",
+        "properties": {
+          "reference_date": {
+            "oneOf": [
+              {
+                "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/value_sources"
+              },
```
MebinAbraham, line 19: For the reference day, we need to validate that when an `answer` source is used, it is one of our date input answer types, i.e reference a text field answer should be invalid but would currently pass validation.

tests/schemas/valid/test_placeholder_transform.json
```diff
@@ -107,6 +107,31 @@
...
+                            "placeholder": "format_date_range_pair",
+                            "transforms": [
+                              {
+                                "transform": "format_date_range_pair",
+                                "arguments": {
+                                  "date_range": ["2021-09-10", "2021-09-15"]
+                                }
+                              }
+                            ]
                           }
```
MebinAbraham, line 135: The schema should define a realistic schema, as we would never have this at run time, we should use a value source of the `reference_date` and 2 values sources for the format date range.

schemas/string_interpolation/transforms/format_date_range_pair.json
jamesnunn: There is already the old version of this which is `format_date_range` - should it be superseded or does this need a new name?

schemas/string_interpolation/transforms/generate_date_range.json
jamesnunn: yeah this is what tripped me up and I was [asking](https://surveydatacollection.slack.com/archives/CLR17KQKU/p1633009031014500) what this section does - I'm still not clear then so I'll catch up about it in a bit

schemas/string_interpolation/transforms/generate_date_range.json
jamesnunn: `date_bounds`?

schemas/string_interpolation/transforms/generate_date_range.json
jamesnunn: I prefer full name, and also I agree constants should be caps but that usually refers the variable name, but fine to change these to caps too. Does what's in the proposal affect this card? If so, should that be completed first and referenced here?

schemas/string_interpolation/transforms/format_date_range_pair.json
MebinAbraham: So it doesn't look like we have a transform `format_date_range` in runner, just validator 🤔 . Given the function that passes input into this has to be a list of inputs, I guess this should supersede the existing `format_date_range`.

schemas/string_interpolation/transforms/generate_date_range.json
MebinAbraham: This is the literal "name" of the transform. `enum` was used in the old JSON draft, since the latest draft `enum` could be dropped for `const` when we only have one value, so instead of enum, it would be like `"const": "format_date_range"` etc.

schemas/string_interpolation/transforms/generate_date_range.json
MebinAbraham: Not sure if that is very clear. Don't mind that, maybe see what others think. Maybe `range` would help, even if bounds implies it is a range, i.e `date_range_bounds`?. If we chose `date_bounds` or `date_range_bounds`, the format transform should be updated accordingly, i.e `format_date_bounds` etc.

schemas/string_interpolation/transforms/generate_date_range.json
MebinAbraham: The proposal isn't for placeholder but we should be consistent, we don't want one place of runner referencing week days as `MON`, and the other as `monday`. I am happy with the full name, and if you are happy with caps, we should change these to `MONDAY` etc, and I can update the proposal to match.

schemas/string_interpolation/transforms/generate_date_range.json
jamesnunn: We also have a transform called `calculate_date_difference` - I just followed existing naming patterns where I could

jamesnunn: Yeah `date_range_bounds` is most descriptive, it's the boundary dates for a date range.

schemas/string_interpolation/transforms/generate_date_range.json
jamesnunn: What would be the alternative - isn't the `started_at` date destined to be a valuesource?

tests/schemas/valid/test_placeholder_transform.json
jamesnunn: I see, I was again using existing patterns as a guide - just above this section is the `email_links` transform which hardcodes some arguments. Is that a different use case? If so, should I be doing this:
```
"reference_date": {
                "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/value_sources"
              }
```

schemas/string_interpolation/transforms/generate_date_range.json
MebinAbraham: Yes `started_at` will be a value source, but I don't follow how your comment relates to my initial comment?. My suggestion was if we support value sources for offset full weeks and days in range, they will need to be checked to ensure type is as expected, as you wouldn't want an text field answer value source for days in range. Given, we only have the requirement to make `reference_date` dynamic, these could be simplified and set to `integer` only which we can then expand if there is required to drive the offsets etc from a dynamic source.

tests/schemas/valid/test_placeholder_transform.json
jamesnunn: Never mind, I found another example to use for this 👍

schemas/string_interpolation/transforms/generate_date_range.json
jamesnunn: From our convo this is out of scope for this one? Did you intend to raise/raise a card?

schemas/string_interpolation/transforms/format_date_range.json
```diff
@@ -11,15 +12,22 @@
...
         "properties": {
-          "start_date": {
-            "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/value_sources"
-          },
-          "end_date": {
-            "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/value_sources"
+          "date_range": {
+            "oneOf": [
+              {
+                "type": "array",
```
MebinAbraham, line 18: We should specify the length as 2. https://json-schema.org/understanding-json-schema/reference/array.html#length

MebinAbraham (commented):
PR title/description needs updating.

schemas/string_interpolation/transforms/generate_date_range.json
MebinAbraham: Yup, a card has been raised to address that.

MebinAbraham (approved):

PR #116, opened 2021-09-30
Proposal for Dynamic answer options

MebinAbraham:
### PR Context
Proposes the structure to support dynamic answer options.

### Checklist

### DIFF
--- a/doc/decisions/0010-dynamic-answer-options.md
+++ b/doc/decisions/0010-dynamic-answer-options.md
@@ -1,0 +1,252 @@
+# 10. Dynamic Answer Options
+
+## Context
+
+We need the ability to generate answer options for Checkbox and Radio answers dynamically.
+
+We need to support dynamic options that are driven by:
+
+- Checkbox answers
+- List collectors
+- Arbitrary functions that return a sequence of values.
+
+Additional requirements:
+
+- Ability for the label and value to be different
+    - When driven by list collectors, the option's `value` is the `list_item_id` and the `label` is some transformed value on answers stored against the list item id, e.g. `FirstName LastName`.
+        - When driven by answers, the option's `value` should be the answer value, and the `label` should be resolved to the schema label based on the answer id, list item id, and option value.
+        - Supports translated schemas since we are resolving to the label.
+        - Allows for value to be different to the label.
+- Values to resolve correctly when inside and outside of repeats.
+
+## Proposal
+
+Introduce a new `dynamic_options` object that will handle the generation of any dynamic options.
+Existing `options` property to stay as is to provide the ability to define additional options. The additional options will always be appended to the bottom.
+
+Properties and uses:
+
+- `dynamic_options.values`
+    - Used to define the literal values used for [`options.*.value`](https://github.com/ONSdigital/eq-questionnaire-validator/blob/master/schemas/answers/definitions.json#L17-L19).
+    - Structure to allow either a Value Source that returns a sequence or the new Rules structure.
+- `dynamic_options.transform`
+    - Used to apply some transform on each value provided to `dynamic_options.values` which will be used for [`options.*.label`](https://github.com/ONSdigital/eq-questionnaire-validator/blob/master/schemas/answers/definitions.json#L14-L16).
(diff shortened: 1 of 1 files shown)


petechd (approved):

doc/decisions/0010-dynamic-answer-options.md
```diff
@@ -0,0 +1,256 @@
...
+                  "days": -7,
+                  "day_of_week": "MONDAY"
+                }
+              ]
+            },
+            7
+          ]
+        },
+        {
+          "format_date": [
```
pricem14pc, line 67: Can the transform not be the argument for the map, rather than the map having a separate operation that simply references the transform? Wouldn't this always resolve to a transform? It stood as as duplication.

MebinAbraham: The transform isn't tied to the map. And it wouldn't always be the same. The `format_date` for map is to return date as `YYYY-MM-DD` for each of the value but the `format_date` in `transform` is for the label which will return as `Monday 11 October 2021` etc. We could allow `transform` to be optional for cases where no transformation is needed.

MebinAbraham: I have updated content to capture that `transforms` can be optional if no transform is needed.

pricem14pc: Yep, I see that now. Thanks @MebinAbraham

doc/decisions/0010-dynamic-answer-options.md
```diff
@@ -0,0 +1,256 @@
...
+            "%Y-%m-%d"
+          ]
+        }
+      ]
+    },
+    "transform": {
+      "format_date": [
+        "self",
+        "%A %d %B %Y",
+        "%Y-%m-%d"
```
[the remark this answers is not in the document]
pricem14pc: Yeah, I'd misread..
> It takes three arguments, the datetime object, or a string date, the format to return and an input format (Used with string inputs).

..thinking the 2nd argument was meant for the string date. Again makes sense now.

doc/decisions/0010-dynamic-answer-options.md
```diff
@@ -0,0 +1,256 @@
...
+      "label": "None of the above",
+      "value": "None of the above"
+    }
+  ]
+}
+```
+
+- The `concat` operator is responsible for concatenating a sequence of values using a delimiter.
+  This example demonstrates formatting the answer values for each list item.
+- When dynamic options are used with a `list` source, the `list_item_id` is not passed explicitly as `self` as the call is responsible for using the rule evaluator in the context of the current list item id.
```
pricem14pc, line 246: This isn't completely clear to me..

> as the call is responsible for using the rule evaluator in the context of the current list item id

..does this mean the `concat` transform itself has the logic to retrieve the `first-name` and `last-name` from the answer store?

pricem14pc (approved):
I think in general this approach is consistent with the direction the patterns in Runner has taken. While some of the less well defined business usecases have been included, the current ones we need this solution to deliver appear sound. We can always revisit this (or raise future proposals) for the less well understood requirements included here.

LJBabbage (approved):

MebinAbraham:
> * Why is `self` needed - isn't the index of an iteration implicit - why couldn't it be worked out in the code and kept out of the schema? It obviously _is_ needed because you spent the time working it out, I just don't get it of course.

The `self` is needed because the operators are not coupled with the caller and no other context than the arguments are passed to it. `self` could be any arbitrary value which is valid to an operator, it isn't the index. Since the operators are independent of the caller, it needs to pass resolved values which operators can then just evaluate. The new rules schema I am conforming to was designed to allow us to recursively nest rules for more advanced usages, which wouldn't be possible if the operators were coupled to the parent caller.

---

> * Is the transform `format_date` the same as the the map function `format_date`? If so, it isn't clear to the uninitiated that it can be, because transforms are named as such, while the map functions aren't identified as such in the code (are they?)

Yes, the transform must conform to the new rules structure, hence the same set of operators too. So in our case, yes the two `format_date` in `map` and `transform` are identical. 

Could you elaborate on comment below, I don't quite follow.
> If so, it isn't clear to the uninitiated that it can be, because transforms are named as such, while the map functions aren't identified as such in the code (are they?)

---

> An example of the steps the data takes from user input, to what's presented back to the user would help a lot.

I'll add in an example of how a resolved `values`/`transform` would look.

PR #119, opened 2021-10-07
Introduce response metadata as value source

petechd:
This adds `response_metadata` as a value source for transforms and  fixes all the affected definitions. New valid test schema added and new validation of the new source, full description on [this Trello card](https://trello.com/c/18e2x6jf).

### How to review
Run `make test`. Check new validator code. In new valid test schema `format_date` transform was used as the new dynamic dates placeholder transforms are not yet ready).

### Checklist

### DIFF
--- a/app/validators/sections/section_validator.py
+++ b/app/validators/sections/section_validator.py
@@ -6,6 +6,7 @@ from app.validators.blocks import get_block_validator
 from app.validators.questionnaire_schema import get_object_containing_key
 from app.validators.questions import get_question_validator
 from app.validators.routing.new_routing_validator import NewRoutingValidator
+from app.validators.routing.new_when_rule_validator import NewWhenRuleValidator
 from app.validators.routing.routing_validator import RoutingValidator
 from app.validators.routing.when_rule_validator import WhenRuleValidator
 from app.validators.validator import Validator
@@ -50,6 +51,12 @@ class SectionValidator(Validator):
             )
             self.errors += when_validator.validate()
 
+    def validate_new_skip_conditions(self, skip_condition, origin_id):
+        when_validator = NewWhenRuleValidator(
+            skip_condition["when"], origin_id, self.questionnaire_schema
+        )
+        self.errors += when_validator.validate()
+
     def validate_value_sources(self):
         source_references = get_object_containing_key(self.section, "identifier")
         for json_path, source_reference in source_references:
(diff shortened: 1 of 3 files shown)


tests/schemas/valid/test_response_metadata_value_source.json
```diff
@@ -0,0 +1,113 @@
+{
```
MebinAbraham, line 1: Not a blocker but this could be added to one of the existing schemas like `test_string_transforms.json`.

tests/schemas/valid/test_response_metadata_value_source.json
```diff
@@ -0,0 +1,113 @@
...
+                      "description": "Submit the answer to create submitted_at identifier for response metadata value source"
+                    }
+                  ]
+                }
+              },
+              "routing_rules": [
+                {
+                  "block": "interstitial"
+                }
+              ]
```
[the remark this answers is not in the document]
petechd: This test schema is now deleted, response metadata transform moved to existing schema, check below.

tests/schemas/valid/test_response_metadata_value_source.json
petechd: Moved response metadata transform to `test_string_transforms.json` and deleted the new schema, thanks!

app/validators/value_source_validator.py
```diff
@@ -49,6 +53,13 @@ def _validate_list_source_reference(self, identifiers):
             if identifier not in self.questionnaire_schema.list_names:
                 self.add_error(self.LIST_REFERENCE_INVALID, identifier=identifier)
 
+    def _validate_response_metadata_source_reference(self, identifiers):
+        for identifier in identifiers:
+            if identifier not in ["started_at"]:
```
LJBabbage, line 58: This looks a bit odd, my feeling is this has been done as you see response_metadata_source expanding, which is fine, however if that was the case I would expect it to be a constant REPSONSE_METADATA_IDENITIFERS. On the other hand I wonder if we have any other use case above started_at, if not then this should be a straight equals and extend when needed

LJBabbage (changes requested):

app/validators/value_source_validator.py
petechd: At the moment only `started_at` as far as I know, should I make it that one identifier only without a list or create a constant with one element for now?

PR #120, opened 2021-10-11
Add validation for skip conditions using new rules design

petechd:
### PR Context
This adds new rules design validator support  for skip conditions. New valid test schema and new skip conditions definition added. Validator differentiates between new when rule validator and old style one based on type of when rule (list vs dict).
Changes are described on [this Trello card](https://trello.com/c/T48XR1at).

### Checklist

### DIFF
--- a/tests/schemas/valid/test_valid_skip_conditions.json
+++ b/tests/schemas/valid/test_valid_skip_conditions.json
@@ -46,7 +46,7 @@
                 "answers": [
                   {
                     "id": "do-you-want-to-skip-answer",
-                    "label": "Do you want to skip?",
+                    "label": "Do you want to skip the interstitial page?",
                     "mandatory": true,
                     "options": [
                       {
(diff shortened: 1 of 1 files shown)


petechd:
[Validator support new rules design for skip conditions (3 SP)](https://trello.com/c/T48XR1at/4902-validator-support-new-rules-design-for-skip-conditions-3-sp)

app/validators/sections/section_validator.py
```diff
@@ -93,9 +101,14 @@ def validate_routing(self, schema_element, group):
...
-            self.validate_skip_conditions(
-                schema_element["skip_conditions"], schema_element["id"]
-            )
+            if isinstance(schema_element["skip_conditions"], list):
+                self.validate_skip_conditions(
+                    schema_element["skip_conditions"], schema_element["id"]
+                )
+            elif isinstance(schema_element["skip_conditions"], dict):
+                self.validate_new_skip_conditions(
+                    [schema_element["skip_conditions"]], schema_element["id"]
```
MebinAbraham, line 110: How come this is passed as a list?. You just need to pass the single when condition.

schemas/common_definitions.json
```diff
@@ -96,6 +96,32 @@
...
+    "items": {
+      "type": "object",
+      "properties": {
+        "when": {
+          "$ref": "https://eq.ons.gov.uk/rules/rule.json"
+        }
+      },
+      "additionalProperties": false
+    },
+    "required": ["when"]
```
MebinAbraham, line 113: "[items](https://json-schema.org/understanding-json-schema/reference/array.html#items)" are used for arrays, our new struct is an object, therefore, this would result in some incorrect validation, for example, this defenition would pass validation as it isn't able to check the rule properly:.
```
  "skip_conditions": {
    "when": {
      "==": [

      ]
    }
  }
```

You need to remove the items check and move the proprieties up a level.

```suggestion
    "properties": {
      "when": {
        "$ref": "https://eq.ons.gov.uk/rules/rule.json"
      }
    },
    "additionalProperties": false,
    "required": ["when"]
```

PR #123, opened 2021-11-03
Add new rules design for section enabled rules

petechd:
### PR Context
This adds new rules engine validation support to section enabled schema element (as well as missing validation of the old style rules for section enabled). Validator is currently not failing on incorrect referenced answer when used in section enabled "when" rule, fix being worked on, check this card: https://trello.com/c/irPw0tnc/4936-update-newwhenrulevalidator-to-check-for-non-existing-references. Test schema with valid section enabled was added.

### Checklist

### DIFF
--- a/app/validators/sections/section_validator.py
+++ b/app/validators/sections/section_validator.py
@@ -44,16 +44,16 @@ class SectionValidator(Validator):
     def validate_section_enabled(self):
         section_enabled = self.section.get("enabled", None)
 
-        if section_enabled and isinstance(section_enabled, list):
-            for enabled in self.section.get("enabled"):
+        if isinstance(section_enabled, list):
+            for enabled in section_enabled:
                 when = enabled["when"]
                 when_validator = WhenRuleValidator(
                     when, self.section["id"], self.questionnaire_schema
                 )
                 self.errors += when_validator.validate()
 
-        if section_enabled and isinstance(section_enabled, dict):
-            when = self.section.get("enabled")["when"]
+        if isinstance(section_enabled, dict):
+            when = section_enabled["when"]
             when_validator = NewWhenRuleValidator(
                 when, self.section["id"], self.questionnaire_schema
             )
--- a/app/validators/sections/section_validator.py
+++ b/app/validators/sections/section_validator.py
@@ -25,6 +25,7 @@ class SectionValidator(Validator):
         self.validate_summary()
         self.validate_groups()
         self.validate_value_sources()
+        self.validate_section_enabled()
         return self.errors
 
     def validate_repeat(self):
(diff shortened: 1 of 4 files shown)


petechd:
[Validator to support new rules design for section enabled rules (3 SP)](https://trello.com/c/fm4zyiDJ/4947-validator-to-support-new-rules-design-for-section-enabled-rules-3-sp)

berroar (approved):
LGTM, tested this and the new schema is validating correctly

app/validators/sections/section_validator.py
```diff
@@ -40,6 +41,24 @@ def validate_summary(self):
...
+
+        if section_enabled and isinstance(section_enabled, list):
+            for enabled in self.section.get("enabled"):
+                when = enabled["when"]
+                when_validator = WhenRuleValidator(
+                    when, self.section["id"], self.questionnaire_schema
+                )
+                self.errors += when_validator.validate()
+
+        if section_enabled and isinstance(section_enabled, dict):
```
MebinAbraham, line 55: You don't really need the `if section_enabled` bit as the `isinstance` would handle it.

app/validators/sections/section_validator.py
```diff
@@ -40,6 +41,24 @@ def validate_summary(self):
             for item in section_summary.get("items", []):
                 self.validate_list_exists(item.get("for_list"))
 
+    def validate_section_enabled(self):
+        section_enabled = self.section.get("enabled", None)
+
+        if section_enabled and isinstance(section_enabled, list):
+            for enabled in self.section.get("enabled"):
```
MebinAbraham, line 48: Worth reusing the already existing var `section_enabled` instead of access again.

app/validators/sections/section_validator.py
```diff
@@ -40,6 +41,24 @@ def validate_summary(self):
...
+                when = enabled["when"]
+                when_validator = WhenRuleValidator(
+                    when, self.section["id"], self.questionnaire_schema
+                )
+                self.errors += when_validator.validate()
+
+        if section_enabled and isinstance(section_enabled, dict):
+            when = self.section.get("enabled")["when"]
```
MebinAbraham, line 56: Worth reusing the already existing var `section_enabled` instead of access again.

app/validators/sections/section_validator.py
petechd: Removed that `section_enabled` if part, thanks!

app/validators/sections/section_validator.py
petechd: Using the var now, thanks!

app/validators/sections/section_validator.py
petechd: Same, using the var, thanks!

app/validators/sections/section_validator.py
```diff
@@ -40,6 +41,24 @@ def validate_summary(self):
...
+
+        if isinstance(section_enabled, list):
+            for enabled in section_enabled:
+                when = enabled["when"]
+                when_validator = WhenRuleValidator(
+                    when, self.section["id"], self.questionnaire_schema
+                )
+                self.errors += when_validator.validate()
+
+        if isinstance(section_enabled, dict):
```
MebinAbraham, line 55:
```suggestion
        elif isinstance(section_enabled, dict):
```

MebinAbraham (approved):

PR #124, opened 2021-11-04
Update value source definition

parimooa:
### PR Context

- Move `lookup_source` schema from `definitions.json` to `value_source.json` schema file to be consistent.
- Update schema properties especially the `list_item_selector` 
- Update tests and schemas
- Update type return default Array type for list when selector is not present

### How to test
Run all tests and ensure no tests are broken.

### Checklist


schemas/value_sources.json
```diff
@@ -38,7 +54,7 @@
     "type": "object",
     "properties": {
       "identifier": {
-        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/identifier",
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/metadata_identifier",
```
MebinAbraham, line 57: 👍

MebinAbraham (commented):
Looks good.

app/validators/routing/types.py
```diff
@@ -51,7 +51,9 @@ def resolve_value_source_json_type(value_source, answers_with_context):
         return ANSWER_TYPE_TO_JSON_TYPE[answer_type]
 
     if source == "list":
-        return LIST_SELECTOR_TO_JSON_TYPE[value_source["selector"]]
+        if selector := value_source.get("selector"):
+            return LIST_SELECTOR_TO_JSON_TYPE[selector]
+        return TYPE_ARRAY
```
MebinAbraham, line 56: We should add a test for this scenario. You can update the existing test named `test_validate_value_sources`. It tests with a selector, just need to add without a selector.

LJBabbage (approved):
LGTM, thought it was a big PR, but was just a few minor tweaks

app/validators/routing/types.py
MebinAbraham: This is still outstanding, the test you added is valid, but this should also test with the full source in `test_validate_value_sources` but without a selector.

MebinAbraham (approved):

PR #125, opened 2021-11-04
Add map, date-range and format-date operators

MebinAbraham:
### PR Context
- Initial pass of https://github.com/ONSdigital/eq-questionnaire-validator/blob/master/doc/decisions/0010-dynamic-answer-options.md with some changes in preparation to support `dynamic_options`.
- Added the `map`, `date-range` and `format-date` operators. Added the new property `day_of_week` to the `date` operator.
- The `NewWhenRuleValidator` has been refactored, and it is now only responsible for validation specific to routing; the newly introduced `RulesValidator` is responsible for the rest. This was done since rules will now be used elsewhere than just when rules.
- Refactored some of the tests since `Pylint` was complaining about duplicate code. Tests from the existing `NewWhenRuleValidator` have been moved to `RulesValidator` where appropriate.
- The top-level JSON schema definition for the new rules have been split into 2. One for the boolean rules used for when rules and one for value rules which will be first used in `dynamic_options`.
- Since `map` currently only supports `format-date` with `date-range`, I have not added any guards to validate values etc. However, the `map` operator is validated when referencing the value `self`. This validation will be reused in `dynamic_options.transform` as `self` cannot be used outside those contexts.

Changes from the initial proposal:
- `format-date` will be used for any date format transformation. It will always require a datetime object; therefore, it accepts the `date` operators, allowing casting string dates to datetime objects. This also means that the `date` operator now supports `self` reference.
- date formats themselves have changed since the formatting of dates are done using flask babel; hence you cannot use Python date formats such as `%Y-%m-%d` etc.

**Ensure the changes are appropriate and no required validation has been missed.**
The changes required for `dynamic_options` will build on this piece of work and will be delivered in a separate PR.

### Checklist


tests/schemas/rules/boolean/value/count.json
```diff
@@ -0,0 +1,98 @@
+[
+  {
+    "description": "A checkbox selected answer equal to number of checkboxes count number",
```
LJBabbage, line 3: This needs to be clearly as I don't really understand what a checkboxes count number is.

app/validators/rules/rule_validator.py
```diff
@@ -0,0 +1,207 @@
...
+    Operator.DATE_RANGE,
+    Operator.MAP,
+]
+
+ALL_OPERATORS = (
+    LOGIC_OPERATORS + COMPARISON_OPERATORS + ARRAY_OPERATORS + VALUE_OPERATORS
+)
+
+
+SELF = "self"
```
LJBabbage, line 50: Self is a bit awkward when used with python instances. I can just about follow it, but my brain did get muddled if it was the class instance or the answer etc

app/validators/rules/rule_validator.py
```diff
@@ -0,0 +1,207 @@
...
+            not in ["Date", "MonthYearDate", "YearDate"]
+        ):
+            self.add_error(
+                self.DATE_OPERATOR_REFERENCES_NON_DATE_ANSWER,
+                value_source=first_argument,
+            )
+
+    def _validate_count_operator(self, operator):
+        """
+        Validates that when an answer value source is used it is a checkbox
```
LJBabbage, line 170: The doc string and name of function don't really go together

schemas/rules/boolean_rule.json
```diff
@@ -1,7 +1,7 @@
 {
-  "$id": "https://eq.ons.gov.uk/rules/rule.json",
+  "$id": "https://eq.ons.gov.uk/rules/boolean_rule.json",
```
LJBabbage, line 2: You're right, all our rules are boolean and this is explicit, but because we will never have a rule that isn't I think could have skipped these changes and reduced the noise

app/validators/rules/rule_validator.py
```diff
@@ -0,0 +1,207 @@
...
+            function_to_map_over_arguments
+        )
+
+        if SELF not in arguments_for_non_map_operators:
+            self.add_error(
+                self.MAP_OPERATOR_WITHOUT_SELF_REFERENCE, rule=operator["map"][0]
+            )
+
+    def _get_flattened_arguments_for_non_map_operators(self, arguments):
+        non_operator_arguments = []
```
LJBabbage, line 134: We seem inconsistent with doc strings in this class

tests/schemas/rules/boolean/value/date.json
```diff
@@ -0,0 +1,177 @@
...
+        {
+          "date": [
+            {
+              "source": "answers",
+              "identifier": "answer-1"
+            }
+          ]
+        },
+        {
+          "date": ["now"]
```
LJBabbage, line 33: Can we have anything other than now as a dynamic date?

LJBabbage (approved):
I am not overly confident with this review, however it seems ok and happy for it to go in. I think we need to try to avoid unnecessary refactoring where possible, although it has made the code better, it has added noise to this PR, which wasn't straight forward to begin with.  There are a few descriptions and terms which I personally found confusing but won't block

tests/schemas/rules/boolean/value/count.json
MebinAbraham: From the existing file, but have updated to be meaningful.

app/validators/rules/rule_validator.py
MebinAbraham: Didn't clock it since it was existing but have made a slight change for clarity.

app/validators/rules/rule_validator.py
MebinAbraham: Have added a doc string for this.

tests/schemas/rules/boolean/value/date.json
MebinAbraham: No, `now` or an actual data via a source or `self` which would be resolved to a string date being calling the date operator.

app/validators/rules/rule_validator.py
MebinAbraham: I could rename the const itself to like `SELF_REFERENCE_KEY` or something like that if it is confusing. Class instance would never be upper case so in that sense I think it is okay but not fussed, so happy to rename for clarity if required.

app/validators/rules/rule_validator.py
```diff
@@ -0,0 +1,215 @@
...
+        "Count operator references non Checkbox answer"
+    )
+    MAP_OPERATOR_WITHOUT_SELF_REFERENCE = (
+        f"Argument one of the `map` operator does not reference `{SELF_REFERENCE_KEY}`"
+    )
+    SELF_REFERENCE_OUTSIDE_MAP_OPERATOR = (
+        f"Reference to {SELF_REFERENCE_KEY} was made outside of the `map` operator"
+    )
+
+    def __init__(self, rules, origin_id, questionnaire_schema):
```
jamesnunn, line 68: Is it worth adding type hints to this new/retouched code? Or was that just something for runner?

MebinAbraham: Ya, we don't use `mypy` on this repo. Not as useful here atm, as we got bigger things to prob consider for validator such as using AJV/upgrading Json schema drafts etc.

app/validators/rules/rule_validator.py
```diff
@@ -0,0 +1,215 @@
...
+        """
+        values = []
+        option_values = []
+        for argument in rules[operator_name]:
+            if isinstance(argument, dict) and argument.get("source") == "answers":
+                option_values = self.questionnaire_schema.answer_id_to_option_values_map.get(
+                    argument["identifier"]
+                )
+            else:
+                values = argument if isinstance(argument, list) else [argument]
```
parimooa, line 206: `option_values` and `values` will only have last values of `argument` ? as they are getting overwritten

app/validators/rules/rule_validator.py
```diff
@@ -0,0 +1,215 @@
...
+            self.add_error(self.SELF_REFERENCE_OUTSIDE_MAP_OPERATOR, rule=rules)
+
+    def _validate_map_operator(self, operator):
+        """
+        Validates that the first argument to the map operator has a reference to `self`
+        The second argument is currently not validated here as it can currently only be `date-range`
+        """
+        function_to_map_over_arguments = list(operator["map"][0].values())[0]
+        if SELF_REFERENCE_KEY in function_to_map_over_arguments:
+            return None
```
parimooa, line 125: Missing test for this statement

parimooa (commented):
Looks Good , few  minor comments

app/validators/rules/rule_validator.py
```diff
@@ -0,0 +1,215 @@
...
+                        operands
+                    )
+            else:
+                non_operator_arguments.append(argument)
+
+        return non_operator_arguments
+
+    def _validate_date_operator(self, operator):
+        """
+        Validates that when an answer value source is used it is a date
```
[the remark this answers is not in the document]
MebinAbraham: Don't think that suggestion quite makes sense?. I added a comma before `it is a date` which should help.

app/validators/rules/rule_validator.py
```diff
@@ -0,0 +1,215 @@
...
+                operator in argument
+                for operator in ALL_OPERATORS
+                if operator != Operator.MAP
+            ):
+                for operands in argument.values():
+                    non_operator_arguments += self._get_flattened_arguments_for_non_map_operators(
+                        operands
+                    )
+            else:
+                non_operator_arguments.append(argument)
```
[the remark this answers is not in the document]
MebinAbraham: Any arguments that are passed to rules which aren't literal operators, i.e either value sources or constants.

app/validators/rules/rule_validator.py
```diff
@@ -0,0 +1,215 @@
...
+        for argument in arguments:
+            if isinstance(argument, dict) and any(
+                operator in argument
+                for operator in ALL_OPERATORS
+                if operator != Operator.MAP
+            ):
+                for operands in argument.values():
+                    non_operator_arguments += self._get_flattened_arguments_for_non_map_operators(
+                        operands
+                    )
```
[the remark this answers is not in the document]
MebinAbraham: Have split these into 2 methods.

app/validators/rules/rule_validator.py
MebinAbraham: This was copied from the existing file, but I see what you mean. However, this works because 2 operators can never be at the same level, it always has to be nested if you have more than 1 operator. So technically  this is always a dict of length one, which can have 2 arguments, i.e:
```py
{'==': [{'source': 'answers', 'identifier': 'string-answer'}, 'Maybe']}
```

This struct would be invalid:
```py
{
	'==': [{'source': 'answers', 'identifier': 'string-answer'}, 'Maybe'],
	'!=': [{'source': 'answers', 'identifier': 'string-answer'}, 'Maybe'],
}
```

`values` should never be populated when the `else` clause is reached if the schema is valid.

.pylintrc
```diff
@@ -249,7 +249,7 @@ notes=FIXME,XXX,TODO
 [SIMILARITIES]
 
 # Minimum lines number of a similarity.
-min-similarity-lines=4
+min-similarity-lines=6
```
MebinAbraham, line 252: This was increased slightly as tests are prone to duplicate code, even with some refactoring there is likely to still be duplicates. Not really worth extra refactors for the sake of having constant for test data between tests which isn't great for readability. Unfortunately, the check cannot be disabled per line/file either. https://github.com/PyCQA/pylint/issues/214

parimooa (approved):

jamesnunn (approved):

PR #126, opened 2021-11-09
Add support for dynamic answer options

MebinAbraham:
### PR Context
- Added remaining implementations defined in [master/doc/decisions/0010-dynamic-answer-options.md](https://github.com/ONSdigital/eq-questionnaire-validator/blob/master/doc/decisions/0010-dynamic-answer-options.md) that wasn't implemented in https://github.com/ONSdigital/eq-questionnaire-validator/pull/125.
- Added support for `dynamic_options` for answers that support `options`.
	- Supports checkbox driven, list collector driven and function driven (date).
- Added the `concat` and `option-label-from-value` operators.
- Answer options length validation is now done in code as `options` are not mandatory when `dynamic_options` are defined.

Note: No validation has been added to ensure that some routing/skip rules exists somewhere in the schema to prevent cases where dynamic options is empty/invalid. The overhead in this is too great. Have confirmed this with Donna.

Ensure the changes are appropriate and no required validation has been missed.

### Checklist

### DIFF
--- a/app/validators/answers/option_answer_validator.py
+++ b/app/validators/answers/option_answer_validator.py
@@ -16,6 +16,9 @@ class OptionAnswerValidator(AnswerValidator):
         "{answer_type} requires at least {required_num_options} answer option(s) but got {actual_num_options}"
     )
     OPTIONS_DEFINED_BUT_EMPTY = "Answer options defined, but empty"
+    DYNAMIC_OPTIONS_REFERENCES_NON_CHECKBOX_ANSWER = (
+        "Dynamic options references non Checkbox answer"
+    )
 
     MIN_OPTIONS_BY_ANSWER_TYPE = {"Checkbox": 1, "Radio": 2, "Dropdown": 2}
 
@@ -104,6 +107,10 @@ class OptionAnswerValidator(AnswerValidator):
         if not self.dynamic_options:
             return None
 
+        self._validate_dynamic_options_answer_source()
+        self._validate_dynamic_options_value_rules()
+
+    def _validate_dynamic_options_value_rules(self):
         for key_to_validate, allow_self_reference in [
             ("values", False),
             ("transform", True),
@@ -115,3 +122,18 @@ class OptionAnswerValidator(AnswerValidator):
                 allow_self_reference=allow_self_reference,
             )
             self.errors += validator.validate()
+
+    def _validate_dynamic_options_answer_source(self):
+        if "source" not in self.dynamic_options["values"]:
+            return None
+
+        value_source = self.dynamic_options["values"]
+        if (
+            value_source["source"] == "answers"
+            and self.questionnaire_schema.get_answer(value_source["identifier"])["type"]
+            != "Checkbox"
+        ):
+            self.add_error(
+                self.DYNAMIC_OPTIONS_REFERENCES_NON_CHECKBOX_ANSWER,
+                value_source=value_source,
+            )
--- a/app/validators/answers/option_answer_validator.py
+++ b/app/validators/answers/option_answer_validator.py
@@ -1,6 +1,7 @@
 from functools import cached_property
 
 from app.validators.answers import AnswerValidator
+from app.validators.rules.rule_validator import RulesValidator
 
 
 class OptionAnswerValidator(AnswerValidator):
@@ -21,8 +22,9 @@ class OptionAnswerValidator(AnswerValidator):
     def __init__(self, schema_element, questionnaire_schema=None):
         super().__init__(schema_element)
         if questionnaire_schema:
-            self.list_names = questionnaire_schema.list_names
-            self.block_ids = questionnaire_schema.block_ids
+            self.questionnaire_schema = questionnaire_schema
+            self.list_names = self.questionnaire_schema.list_names
+            self.block_ids = self.questionnaire_schema.block_ids
 
     def validate(self):
         super().validate()
@@ -30,6 +32,7 @@ class OptionAnswerValidator(AnswerValidator):
         self.validate_duplicate_options()
         self.validate_labels_and_values_match()
         self.validate_default_exists_in_options()
+        self.validate_dynamic_options()
         return self.errors
 
     @cached_property
@@ -42,7 +45,7 @@ class OptionAnswerValidator(AnswerValidator):
 
     def validate_min_options(self):
         options_len = len(self.options)
-        min_options = self.MIN_OPTIONS_BY_ANSWER_TYPE[self.answer["type"]]
+        min_options = self.MIN_OPTIONS_BY_ANSWER_TYPE[self.answer_type]
 
         if self.dynamic_options:
             if "options" in self.answer and options_len == 0:
@@ -50,7 +53,7 @@ class OptionAnswerValidator(AnswerValidator):
         elif options_len < min_options:
             self.add_error(
                 self.INVALID_NUMBER_OF_ANSWER_OPTIONS.format(
-                    answer_type=self.answer["type"],
+                    answer_type=self.answer_type,
                     required_num_options=min_options,
                     actual_num_options=len(self.options),
                 )
@@ -96,3 +99,19 @@ class OptionAnswerValidator(AnswerValidator):
             option["value"] for option in self.options
         ]:
             self.add_error(self.ANSWER_DEFAULT_MISSING, default_value=default_value)
+
+    def validate_dynamic_options(self):
+        if not self.dynamic_options:
+            return None
+
+        for key_to_validate, allow_self_reference in [
+            ("values", False),
+            ("transform", True),
+        ]:
+            validator = RulesValidator(
+                self.dynamic_options[key_to_validate],
+                self.answer_id,
+                self.questionnaire_schema,
+                allow_self_reference=allow_self_reference,
+            )
+            self.errors += validator.validate()
--- a/app/validators/answers/option_answer_validator.py
+++ b/app/validators/answers/option_answer_validator.py
@@ -10,6 +10,13 @@ class OptionAnswerValidator(AnswerValidator):
     LIST_NAME_MISSING = "List name defined in action params does not exist"
     BLOCK_ID_MISSING = "Block id defined in action params does not exist"
     ANSWER_DEFAULT_MISSING = "Couldn't find matching value for answer default"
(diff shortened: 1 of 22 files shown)


app/validators/answers/option_answer_validator.py
```diff
@@ -10,24 +11,57 @@ class OptionAnswerValidator(AnswerValidator):
...
+    INVALID_NUMBER_OF_ANSWER_OPTIONS = (
+        "Not enough options. "
+        "{answer_type} requires at least {required_num_options} answer option(s) but got {actual_num_options}"
+    )
+    OPTIONS_DEFINED_BUT_EMPTY = "Answer options defined, but empty"
+    DYNAMIC_OPTIONS_REFERENCES_NON_CHECKBOX_ANSWER = (
+        "Dynamic options references non Checkbox answer"
+    )
+
+    MIN_OPTIONS_BY_ANSWER_TYPE = {"Checkbox": 1, "Radio": 2, "Dropdown": 2}
```
parimooa, line 23: With no of errors increasing , i personally prefer to have these error messages have its own class or file. we should really be looking at [Enum](https://docs.python.org/3.11/howto/enum.html)

`MIN_OPTIONS_BY_ANSWER_TYPE` is an odd one out among errors

app/validators/answers/option_answer_validator.py
```diff
@@ -68,3 +102,38 @@ def validate_default_exists_in_options(self):
             option["value"] for option in self.options
         ]:
             self.add_error(self.ANSWER_DEFAULT_MISSING, default_value=default_value)
+
+    def validate_dynamic_options(self):
```
parimooa, line 106: There is no one checking for return value of this method ?  you can just `return`

app/validators/answers/option_answer_validator.py
```diff
@@ -68,3 +102,38 @@ def validate_default_exists_in_options(self):
...
+                self.dynamic_options[key_to_validate],
+                self.answer_id,
+                self.questionnaire_schema,
+                allow_self_reference=allow_self_reference,
+            )
+            self.errors += validator.validate()
+
+    def _validate_dynamic_options_answer_source(self):
+        if "source" not in self.dynamic_options["values"]:
+            return None
```
parimooa, line 128: same as above

app/validators/answers/option_answer_validator.py
```diff
@@ -68,3 +102,38 @@ def validate_default_exists_in_options(self):
...
+
+        value_source = self.dynamic_options["values"]
+        if (
+            value_source["source"] == "answers"
+            and self.questionnaire_schema.get_answer(value_source["identifier"])["type"]
+            != "Checkbox"
```
parimooa, line 134: I don't see  check for `Radio` or `Dropdown` , are they not valid types

schemas/value_sources.json
```diff
@@ -106,5 +106,22 @@
         "$ref": "https://eq.ons.gov.uk/value_sources.json#/location_value_source"
       }
     ]
+  },
+  "any_value_source_except_location": {
+    "description": "Any value source",
+    "anyOf": [
+      {
+        "$ref": "https://eq.ons.gov.uk/value_sources.json#/answer_value_source"
```
parimooa, line 114: `answer_value_source` has `list_item_selector` property which in turn has `location_value_source` ?

schemas/value_sources.json
```diff
@@ -106,5 +106,22 @@
         "$ref": "https://eq.ons.gov.uk/value_sources.json#/location_value_source"
       }
     ]
+  },
+  "any_value_source_except_location": {
+    "description": "Any value source",
```
parimooa, line 111: description is not correct , shouldn't be _Any value source except location_

parimooa (commented):
Looks Ok , few minor comments

tests/schemas/rules/value/map.json
```diff
@@ -26,6 +26,30 @@
       ]
     }
   },
+  {
+    "description": "Map over a list of dates using `date-range` operator using a dynamic date source `now`",
```
LJBabbage, line 30: Map is opposite to what it usually would be in Python. I might have been tempted to explicitly express Map rather than assume knowledge

LJBabbage (commented):
Good stuff, very minor observations

schemas/value_sources.json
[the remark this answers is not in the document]
MebinAbraham: That's fine, as it resolves to an answer.

app/validators/answers/option_answer_validator.py
MebinAbraham: I agree but kept it consistent with what we have. I think it all the validators should be refactored in that way. 

I have made `MIN_OPTIONS_BY_ANSWER_TYPE` a module level constant.

app/validators/answers/option_answer_validator.py
MebinAbraham: It because the validate methods populates the object's errors attribute. It doesn't return the errors. This is consistent with our other validate methods.

app/validators/answers/option_answer_validator.py
MebinAbraham: See above + here, I would mind doing the if check outside the method and calling the method when appropriate. Thoughts?.

app/validators/answers/option_answer_validator.py
MebinAbraham: Only `Checkbox` is a valid type as this is validating the source and only the checkbox answer is a list answer.

schemas/value_sources.json
MebinAbraham: That's fine, as that will still resolve to an answer value.

app/validators/rules/rule_validator.py
```diff
@@ -147,6 +156,16 @@ def _get_flattened_arguments_for_non_map_operators(self, arguments):
 
         return non_operator_arguments
 
+    def _validate_option_label_from_value_operator(self, operator):
+        """
+        Validate the referenced answer id exists
```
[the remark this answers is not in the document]
MebinAbraham: Made a slight change but pretty much similar.

tests/schemas/rules/value/map.json
MebinAbraham: I don't follow. Can you elaborate?.

app/validators/answers/option_answer_validator.py
parimooa: I think i was not clear,  what i meant on line 108 you are returning None , you can drop `None` just use `return`.  If there is no explicit check for `None` there is no point in returning `None`

app/validators/answers/option_answer_validator.py
```diff
@@ -1,6 +1,9 @@
 from functools import cached_property
 
 from app.validators.answers import AnswerValidator
+from app.validators.rules.rule_validator import RulesValidator
+
+MIN_OPTIONS_BY_ANSWER_TYPE = {"Checkbox": 1, "Radio": 2, "Dropdown": 2}
```
jamesnunn, line 6: If these strings are referenced often, I think it'd be better to create and use constants

app/validators/answers/option_answer_validator.py
MebinAbraham: As discussed, this was to return it explicitly. I know we don't have mypy here, but this would be consistent with runner, as implicit optional is disabled there. But I don't mind, either. Do you want me to change?

app/validators/answers/option_answer_validator.py
MebinAbraham: I could make an `AnswersTypes` class and reference the types using that, but not sure if that refactor should be part of this. As it would change references in a lot of unrelated files. Throughts?

jamesnunn: Isn't this new code though? I can't see where it moved from so assumed it was new

MebinAbraham: Yes, that dict is new, but I assume you are referring to having some const to refer to the answer types?. What I mean is, the answer types such as "Checkbox" is referenced in other places that I haven't touched, so I'm not sure if you are suggesting refactor for this file only?

app/validators/answers/option_answer_validator.py
parimooa: should be good as it is if it is consistent with runner

tests/schemas/rules/value/map.json
LJBabbage: I look at it again and it is the right way around

LJBabbage (approved):

app/validators/answers/option_answer_validator.py
MebinAbraham: Added an `AnswerType` Enum class and updated the references in the new code.

jamesnunn (approved):

parimooa (approved):
Looks good , nice use of enums.

MebinAbraham:
@jamesnunn @parimooa @LJBabbage 
Realised schema tests were missing for `concatenate` and also it was still referenced as `concat`. Have added those. So worth having a quick look at [`d8f708d` (#126)](https://github.com/ONSdigital/eq-questionnaire-validator/pull/126/commits/d8f708db17c059a36d8f754170b71541b8646f9e). No extra python validation has been done, since we don't do it for the placeholder concat transform, the overhead is prob not worth it since value sources are pretty loose.

tests/schemas/rules/value/concatenate.json
```diff
@@ -0,0 +1,56 @@
...
+            "source": "answers",
+            "identifier": "last-name"
+          }
+        ],
+        ","
+      ]
+    }
+  },
+  {
+    "description": "Concatenate two metadata",
```
LJBabbage, line 39: minor, but the other 2 you highlighted the separator.

LJBabbage (approved):
Changes look fine. I wonder if we need 2 for answers with a different separators. I think one for answers with a space and one for metadata with comma cover it

parimooa (approved):

tests/schemas/rules/value/concatenate.json
MebinAbraham: It is now one for answers with a space and one for metadata with comma

jamesnunn (approved):

PR #127, opened 2021-11-17
Fix bug with options validator

MebinAbraham:
### PR Context
Small PR to fix a bug in options validator where it was expecting null values to be a valid answer option. Previously passed due to a bug with `answer_id_to_option_values_map` which didn't return the right answer options, but the last merge fixed that. Tests explicitly set `answer_id_to_option_values_map` hence that didn't pick it up. However, we explicitly set property in test when we shouldn't, so that will need revisiting as this is a classic example of why properties should be overridden. We also need to revisit the use of `cached_property` when using generators, as they do not go well together, as a generator can only be consumed once.

Run the validator against runner schemas and ensure they all pass validation.

### Checklist

### DIFF
--- a/app/validators/rules/rule_validator.py
+++ b/app/validators/rules/rule_validator.py
@@ -228,7 +228,7 @@ class RulesValidator(Validator):
 
         if values and option_values:
             for value in values:
-                # Null values are allowed and will not existing in answer options
+                # Null values are allowed and will not exist in answer options
                 if value and value not in option_values:
                     self.add_error(
                         self.VALUE_DOESNT_EXIST_IN_ANSWER_OPTIONS,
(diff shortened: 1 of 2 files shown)


berroar (approved):

app/validators/rules/rule_validator.py
```diff
@@ -228,9 +228,10 @@ def _validate_options(self, rules, operator_name):
 
         if values and option_values:
             for value in values:
-                if value not in option_values:
+                # Null values are allowed and will not existing in answer options
```
MebinAbraham, line 231:
```suggestion
                # Null values are allowed and will not existing in answer options
```
```suggestion
                # Null values are allowed and will not exist in answer options
```

petechd (approved):
Passes local validation, all good!

PR #128, opened 2021-11-23
Use ajv validator

parimooa:
### PR Context

- Replace python Draft7schema validator to node based AJV schema validator
- Create node based Dockerfile For AJV validator
- update workflow to push newly created AJV docker image

[Companion Runner PR ](https://github.com/ONSdigital/eq-questionnaire-runner/pull/724)

Trello card : https://trello.com/c/TUoAGONd/4982-use-ajv-to-validate-json-schema-5-sp
### How to test
Ensure all tests are passing 
Make sure `make run` runs both python and AJV service
Update any valid schema with wrong values and post to `http://localhost:5001/validate` ( post `make run`), ensure the error response is sent with details of error

### Checklist


README.md
```diff
@@ -70,10 +70,10 @@ Also included is a node based version of the json schema validation which may be
 debugging errors. This returns more errors than we'd currently like due to the way polymorphism works for each of our
 blocks.
 
-Run the ajv based version of validator from within the ajv/ directory.
+Run the ajv based version of validator.
 
 ```
-DEBUG=validator node app.js
+yarn start
```
LJBabbage, line 76: The rest of the Readme use make so I was expecting
```
make run-ajv
```

package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "lint": "yarn lint:schemas && yarn lint:test-schemas && yarn lint:ajv",
+    "start": "cd ajv && DEBUG=ajv-schema-validator node app.js &",
```
LJBabbage, line 13: Any reason you used start, it feels like it should be some ajv ish?

parimooa: node apps usually follow convention as `start` for starting server hence the name. Happy to change if required

LJBabbage: I am guessing it does run node app.js :thinking:

parimooa: Yes it does `start` node server

app/views/validate.py
```diff
@@ -1,17 +1,22 @@
...
 from app.validators.questionnaire_validator import QuestionnaireValidator
-from app.validators.schema_validator import SchemaValidator
 
 logger = get_logger()
 
 validate_blueprint = Blueprint("validate", __name__)
 
+AJV_HOST = os.getenv("AJV_HOST", "localhost")
+
+AJV_VALIDATOR_URI = f"http://{AJV_HOST}:5002/validate"
```
MebinAbraham, line 18: Minor but URL since the protocol is defined.
```suggestion
AJV_VALIDATOR_URL = f"http://{AJV_HOST}:5002/validate"
```

app/views/validate.py
```diff
@@ -44,14 +49,18 @@ def validate_schema(data):
...
+        if ajv_response_dict:
+            response["errors"] = ajv_response_dict["errors"]
+            logger.info("Schema validator returned errors", status=400)
+            return jsonify(response), 400
 
-    if len(schema_validator.errors) > 0:
-        response["errors"] = schema_validator.errors
-        logger.info("Schema validator returned errors", status=400)
-        return jsonify(response), 400
+    except requests.exceptions.ConnectionError:
```
MebinAbraham, line 61: Catch all request exceptions?. We don't want to hit the custom python validator if ajv isn't run, which could be other things than connection error.

package.json
MebinAbraham: - Can we make this work without checking out to `ajv` dir?. Seems wrong for a command to change my working dir.
- I don't think we should run it in the background, I think it's fine for a server to be blocking the terminal. Not a big deal, but if we are keeping that, then we should add a stop command and document that.

app/views/validate.py
parimooa: Personally i don't like broad exceptions, i will update it to capture all request exception

package.json
parimooa: If it blocks the terminal then flask can't run. Flask has dependency on AJV service. I will add stop command and update readme

app/views/validate.py
MebinAbraham: Ya, I didn't mean just `Exception` just request acception.

LJBabbage (approved):
Worked for me when posted to endpoint and picked up errors when I broke it

Dockerfile-ajv
```diff
@@ -0,0 +1,16 @@
+FROM node:16
+# Create app directory
+WORKDIR /usr/src/app
+
+# Install app dependencies
+COPY package.json ./
```
MebinAbraham, line 6: I think anything that isn't dev dependencies should be moved from "devDependencies" in package.json. We should then only install non dev dependencies here.

MebinAbraham (commented):
Just that one minor comment, other than that, everything seems to be as expected 👍

Dockerfile-ajv
parimooa: created dependencies in `package.json `and only install required dependencies using 

 `yarn install --production=true`

MebinAbraham (approved):

PR #129, opened 2021-12-01
Validate non existent identifier for value sources

parimooa:
### PR Context

**Background:** Currently the validator throws error for a non existent answer `identifier` in schema. 
  The reason is that,  even though initial check of `Sectionvalidator.validate_groups()` find the error , it still goes ahead and checks again for the type of non existing, as a result of this validator throws the exceptions.

This PR introduced change where validator will first check if value source identifier is valid or not , if it exists then only further checks will be carried out or the validator will return with the error

Related [Runner PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/729)

### How to test
Take any schema and replace identifier with any non existent value. Test for different value sources like `Answers`, `List`, and `metadata`, The validator server should respond with valid error message  without (server) throwing any exceptions
### Checklist


app/validators/value_source_validator.py
```diff
@@ -68,7 +68,7 @@ def _validate_answer_source_reference(self, identifiers, selector=None):
             if identifier not in answers_with_context:
                 self.add_error(self.ANSWER_REFERENCE_INVALID, identifier=identifier)
 
-            if selector:
+            if selector and identifier in answers_with_context:
```
MebinAbraham, line 71: Does the fact that the identifier doesn't exist get picked up later on?. If not, we should raise an error here, as we are otherwise masking the issue.

tests/test_questionnaire_validator.py
```diff
@@ -345,9 +339,10 @@ def test_invalid_relationship_no_list_specified():
...
 
     expected_for_list_error = {
-        "message": error_messages.FOR_LIST_NEVER_POPULATED,
-        "list_name": "not-a-list",
-        "block_id": "relationships",
+        "message": ValueSourceValidator.ANSWER_REFERENCE_INVALID,
+        "identifier": "first-name",
+        "json_path": "groups.[0].blocks.[0].question.title.placeholders.[0].transforms.["
+        "0].arguments.list_to_concatenate.[0].identifier",
     }
```
MebinAbraham, line 346: To me, this implies the schema isn't correct enough to test this scenario?. The body no longer matches the test name.

app/validators/value_source_validator.py
parimooa: `Identifier` doesn't exist gets picked up at line 68.   I don't understand what issue is getting masked? if the `identifier` doesn't exist , you can't get the `answer_type `for `selector` hence subsequent checks for `answer_type` are irrelevant.

tests/test_questionnaire_validator.py
parimooa: I wasn't confident on this, you may be correct. As this json was getting used for other tests so i was reluctant to change it. In this case ,perhaps my lack of knowledge in schema.

app/validators/value_source_validator.py
MebinAbraham: Right, missed line 68. So since that checks if it exists. This condition can be simplified and just be `elif selector:`

tests/test_questionnaire_validator.py
MebinAbraham: I think the JSON has an invalid ID that needs fixing before getting to the bit that does the actual test this intends.

LJBabbage (commented):
The initial change is fine, but I unsure why so much of the JSON was changed, was it all needed? With the test name change I am also thinking that there is a reference to a list that doesn't exist, but I can't see it. Any chance you can point it out for me. Thank you

tests/schemas/invalid/test_invalid_relationship_list_doesnt_exist.json
```diff
@@ -305,8 +471,8 @@
                         }
                       },
                       {
-                        "value": "Legally Registered Civil Partner",
-                        "label": "Legally Registered Civil Partner",
+                        "label": "Legally registered civil partner",
+                        "value": "Legally registered civil partner",
```
MebinAbraham, line 475: Can we reduce the schema by only having 2 options, having all these are not necessary for the test.

parimooa:
> The initial change is fine, but I unsure why so much of the JSON was changed, was it all needed? With the test name change I am also thinking that there is a reference to a list that doesn't exist, but I can't see it. Any chance you can point it out for me. Thank you

The reason is that due to value source check being moved before section check. the existing json didn't have listcollector block hence It throws error as invalid answer reference and dont reach at block level. in order to get past it , json needed to be updated ( include list collector). Also the test name wasn't correct. It is actually checking if the list name exists for that relationship. Hope that makes it clear. The list reference is called `not-a-list` which is non-existent

LJBabbage (approved):

MebinAbraham (approved):

PR #130, opened 2021-12-09
Add new rules support for question_variant and content_variant

LJBabbage:
### PR Context
This PR allows question and content variant to use the new when rules

### Checklist
Check that it validates and looks correct. You can't check the new schemas in runner unfortunately as they are dependent on a future piece of work

N.B You might have noticed I have updated the schemas in this area (although I left ids the same). For the following reason
1. The When was comparing str and int so blew up in runner
2. There was a daft session_timeout_in_seconds (3 seconds) on it
3. The content was garbage, made no sense and didn't reflect how question and content variant are used. For the new JSON I just copied the old one after changes and then updated the when if that makes it easier to review.

### DIFF
--- a/app/validators/sections/section_validator.py
+++ b/app/validators/sections/section_validator.py
@@ -171,10 +171,18 @@ class SectionValidator(Validator):
 
         for variant in all_variants:
             when_clause = variant.get("when", [])
-            when_validator = WhenRuleValidator(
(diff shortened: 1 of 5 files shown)


MebinAbraham (approved):
👍

rmccar (approved):

PR #131, opened 2021-12-13
Adds a schema transform to get the label for an answer value

jamesnunn:
### PR Context
Supports https://github.com/ONSdigital/eq-questionnaire-runner/pull/734

### Checklist

### DIFF
--- a/schemas/string_interpolation/transforms/option_label_from_value.json
+++ b/schemas/string_interpolation/transforms/option_label_from_value.json
@@ -1,0 +1,29 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://eq.ons.gov.uk/string_interpolation/transforms/option_label_from_value.json",
+  "option_label_from_value": {
+    "description": "Get the label of an answer option based on its value. Takes two arguments, the option value and the answer id",
+    "type": "object",
+    "properties": {
+      "transform": {
+        "type": "string",
+        "enum": ["option_label_from_value"]
+      },
+      "arguments": {
+        "type": "object",
+        "properties": {
+          "value": {
+            "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/placeholder_value_sources"
+          },
+          "answer_id": {
+            "type": "string"
+          }
+        },
+        "required": ["value", "answer_id"],
+        "additionalProperties": false
+      }
+    },
+    "additionalProperties": false,
+    "required": ["transform", "arguments"]
+  }
+}
--- a/tests/schemas/valid/test_placeholder_transform.json
+++ b/tests/schemas/valid/test_placeholder_transform.json
@@ -269,6 +269,90 @@
           ]
         }
       ]
+    },
+    {
+      "id": "checkbox-section",
+      "groups": [
+        {
+          "blocks": [
+            {
+              "type": "Question",
+              "id": "mandatory-checkbox",
+              "question": {
+                "answers": [
+                  {
+                    "id": "mandatory-checkbox-answer",
+                    "mandatory": true,
+                    "options": [
+                      {
+                        "label": "Head",
+                        "value": "Head",
+                        "q_code": "0"
+                      },
+                      {
+                        "label": "Body",
+                        "value": "Body",
+                        "q_code": "1"
+                      }
+                    ],
+                    "type": "Checkbox"
+                  }
+                ],
+                "id": "mandatory-checkbox-question",
+                "title": "When you had your accident, where did you sustain injuries?",
+                "type": "General"
+              }
+            },
+            {
+              "type": "Question",
+              "id": "recovery-question-checkbox-block",
+              "question": {
+                "id": "recovery-question",
+                "title": {
+                  "text": "How long did it take to recover from the injury to {body_part}?",
+                  "placeholders": [
+                    {
+                      "placeholder": "body_part",
+                      "transforms": [
+                        {
+                          "transform": "first_non_empty_item",
+                          "arguments": {
+                            "items": [
+                              {
+                                "source": "answers",
+                                "identifier": "mandatory-checkbox-answer"
+                              }
+                            ]
+                          }
+                        },
+                        {
+                          "transform": "option_label_from_value",
+                          "arguments": {
+                            "value": {
+                              "source": "previous_transform"
+                            },
+                            "answer_id": "mandatory-checkbox-answer"
+                          }
+                        }
+                      ]
+                    }
+                  ]
+                },
+                "type": "General",
+                "answers": [
+                  {
+                    "id": "recovery-checkbox-answer",
+                    "label": "Recovery time",
+                    "mandatory": false,
+                    "type": "Number"
+                  }
+                ]
+              }
+            }
+          ],
+          "id": "checkboxes"
+        }
+      ]
     }
   ]
 }
--- a/schemas/string_interpolation/definitions.json
+++ b/schemas/string_interpolation/definitions.json
@@ -126,6 +126,9 @@
             },
             {
               "$ref": "https://eq.ons.gov.uk/string_interpolation/transforms/list_item_count.json#/list_item_count"
+            },
+            {
+              "$ref": "https://eq.ons.gov.uk/string_interpolation/transforms/option_label_from_value.json#/option_label_from_value"
             }
           ]
         }


jamesnunn:
[Using answer option labels as placeholders (3 SP)](https://trello.com/c/YBSWn2q1/4867-using-answer-option-labels-as-placeholders-3-sp)

schemas/string_interpolation/transforms/option_label_from_value.json
```diff
@@ -0,0 +1,29 @@
...
+      },
+      "arguments": {
+        "type": "object",
+        "properties": {
+          "value": {
+            "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/placeholder_value_sources"
+          },
+          "answer_id": {
+            "type": "string"
+          }
```
MebinAbraham, line 20: This answer id should be validated in code to ensure it exists. It is done for rules code, but since this is an arbitrary string we want to ensure it is valid at build time. https://github.com/ONSdigital/eq-questionnaire-validator/blob/b0a937b566b6a172d82159ce0c0366bc57bb56b1/app/validators/rules/rule_validator.py#L187-L195

MebinAbraham: We should also validate that this answer ID also has an answer type which allows answers options.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -60,6 +67,54 @@ def validate_placeholder_object(self, placeholder_object):
...
+                self.add_error(
+                    error_messages.ANSWER_OPTION_LABEL_VALUE_TYPE_INVALID,
+                    identifier=argument,
+                )
+        else:
+            self.add_error(
+                error_messages.ANSWER_OPTION_LABEL_VALUE_TYPE_INVALID,
+                identifier=argument,
+            )
+
```
MebinAbraham, line 117: We just need to check if answer type is Checkbox/Radio/Dropdown. We don't need to check if options exists, cause if it is one of those 3, it will have options, if it didn't it won't get this far.

```py
        if not any(
            x.value == answers[argument]["answer"]["type"] for x in AnswerOptionType
        ):
            self.add_error(
                error_messages.ANSWER_OPTION_LABEL_VALUE_TYPE_INVALID,
                identifier=argument,
            )
```

---

Not sure if we even need `AnswerOptionType` enum class as this is specific to this and there is another answer type (`Relationship`) which has `options` but we don't support here. Could be just a set. Not a blocker though.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -60,6 +67,54 @@ def validate_placeholder_object(self, placeholder_object):
...
+        """
+        Validate answer_id exists in Answer Context
+        Args:
+         argument: answer id
+        Returns:
+            True  : if exists
+            False : if not exists
+        """
+
+        if isinstance(argument, str):
```
MebinAbraham, line 80: Why is this isinstance needed?

app/validators/placeholders/placeholder_validator.py
```diff
@@ -60,6 +67,54 @@ def validate_placeholder_object(self, placeholder_object):
...
+            if argument not in answers:
+                self.add_error(
+                    ValueSourceValidator.ANSWER_REFERENCE_INVALID, identifier=argument
+                )
+                return False
+            return True
+
+    def validate_option_label_value_placeholder(self, argument):
+        """
+        validate answer_id  for transform 'option_label_from_value' is of type of ['Checkbox','Radio','Dropdown']
```
MebinAbraham, line 91: Also checks if answer id exists.

app/error_messages.py
```diff
@@ -4,6 +4,9 @@
 ANSWER_REFERENCE_CANNOT_BE_USED_ON_MIN = (
     "The referenced answer cannot be used to set the minimum of answer"
 )
+ANSWER_OPTION_LABEL_VALUE_TYPE_INVALID = (
+    "The referenced answer_id type is not of type ['Radio','Checkbox','Dropdown']"
```
MebinAbraham, line 8: Given this is specific, it would be good to be a bit more explicit, as it didn't have enough context.  We can also reuse this in rules validator when we come to do that.
```suggestion
    "The referenced answer for option label from value placeholder/operator is not of type ['Radio','Checkbox','Dropdown']"
```

app/validators/placeholders/placeholder_validator.py
parimooa: Actually it is not needed anymore

app/validators/placeholders/placeholder_validator.py
parimooa: i think we need `option` check because if we have valid `answer_id` which doesn't have `options` it will raise key error for `type` ?

MebinAbraham: `type` is a key that exists in every answer.

tests/schemas/valid/test_placeholder_transform.json
```diff
@@ -4,12 +4,9 @@
   "schema_version": "0.0.1",
```
MebinAbraham, line 4: How come this entire schema has changed?. It has other placeholders it was testing.

tests/test_placeholder_validator.py
```diff
@@ -73,3 +81,43 @@ def test_placeholder_plurals():
...
+            ],
+        ),
+    ],
+)
+def test_validation_option_label_from_value(answer_id, expected_error):
+    filename = "schemas/invalid/test_invalid_placeholder_option_label_from_value.json"
+    schema_file = _open_and_load_schema_file(filename)
+    schema = QuestionnaireSchema(schema_file)
+    validator = PlaceholderValidator({})
+    validator.questionnaire_schema = schema
```
MebinAbraham, line 121: The schema should be passed on initialisation instead of assigning it.

tests/test_placeholder_validator.py
```diff
@@ -73,3 +81,43 @@ def test_placeholder_plurals():
     ]
 
     assert validator.errors == expected_errors
+
+
+def _open_and_load_schema_file(file):
+    with open(
+        os.path.join(os.path.dirname(os.path.realpath(__file__)), file), encoding="utf8"
+    ) as json_file:
+        return json.load(json_file)
```
MebinAbraham, line 90: This function is defined in `test_questionnaire_validatory.py`, we should share between the tests.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -60,6 +67,46 @@ def validate_placeholder_object(self, placeholder_object):
...
+        """
+        validate answer_id exists and answer_id for transform 'option_label_from_value' is of type of ['Checkbox','Radio','Dropdown']
+        Args:
+         argument: answer id argument passed to transform
+
+        """
+        answers = self.questionnaire_schema.answers_with_context
+        answer_id_exists = self.validate_placeholder_answer_id(argument)
+
+        # if answer id doesn't exist , no further validation is done
```
MebinAbraham, line 98:
```suggestion
        # if answer id doesn't exist, no further validation is done
```

MebinAbraham (approved):

LJBabbage (changes requested):
Works as expected, a few arbitrary suggestions to make it a slightly easier to follow

app/validators/placeholders/placeholder_validator.py
```diff
@@ -60,6 +67,46 @@ def validate_placeholder_object(self, placeholder_object):
...
+                ValueSourceValidator.ANSWER_REFERENCE_INVALID, identifier=argument
+            )
+            return False
+        return True
+
+    def validate_option_label_value_placeholder(self, argument):
+        """
+        validate answer_id exists and answer_id for transform 'option_label_from_value' is of type of ['Checkbox','Radio','Dropdown']
+        Args:
+         argument: answer id argument passed to transform
```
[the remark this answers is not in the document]
LJBabbage: Weirdly this is below as I look it now, i.e suggestion around answer_id rather than argument

LJBabbage (approved):

tests/schemas/invalid/test_invalid_placeholder_option_label_from_value.json
```diff
@@ -0,0 +1,97 @@
...
+                    "id": "body-part-name",
+                    "label": "Body Part",
+                    "mandatory": true,
+                    "type": "TextField"
+                  }
+                ]
+              }
+            },
+            {
+              "type": "Interstitial",
```
MebinAbraham, line 56: This schema is now invalid. It should only be invalid for the part that we want to be invalid, however, the JSON itself won't pass the validator as `Interstitial` cannot have `question`. To use interstitial, you need to use content block instead of question and pipe the placeholder into the content block's title etc.

MebinAbraham (changes requested):

MebinAbraham (approved):

PR #133, opened 2022-01-19
Fix issue with questions-with-context finding all questions keys

LJBabbage:
### PR Context
At present def questions_with_context is returning all elements in the schema with the key "question", assuming each has a child element "answers" in. However the key "question" can also be used in a content block, at which point it will have no child "answers" and be a str instead throwing a 500. This PR makes sure question is a dict

### Checklist


petechd (approved):

parimooa (approved):

MebinAbraham (approved):
New schema as expected 👍

PR #140, opened 2022-03-03
Add introduction guidance

petechd:
Adds introduction page guidance banner as described on [this trello](https://trello.com/c/Dog3Cv1U). Tests and validator fix added.

### How to review 
New functional test added, do ` yarn test_functional --spec ./tests/functional/spec/introduction.spec.js`.

### DIFF
--- a/schemas/blocks/introduction.json
+++ b/schemas/blocks/introduction.json
@@ -36,7 +36,7 @@
             }
           },
           "additionalProperties": false,
-          "required": ["id", "title", "contents"]
+          "required": ["id", "title"]
         }
       },
       "preview_content": {
--- a/schemas/blocks/introduction.json
+++ b/schemas/blocks/introduction.json
@@ -36,7 +36,18 @@
             }
           },
           "additionalProperties": true,
-          "required": ["id", "title"]
+          "required": ["id"],
+          "anyOf": [
+            {
+              "required": ["title"]
+            },
+            {
+              "required": ["contents"]
+            },
+            {
+              "required": ["guidance"]
+            }
+          ]
         }
       },
       "preview_content": {
--- a/schemas/blocks/introduction.json
+++ b/schemas/blocks/introduction.json
@@ -30,9 +30,6 @@
             },
             "contents": {
               "$ref": "https://eq.ons.gov.uk/common_definitions.json#/contents"
-            },
-            "guidance": {
-              "$ref": "https://eq.ons.gov.uk/common_definitions.json#/guidance"
             }
           },
           "additionalProperties": true,
@@ -43,9 +40,6 @@
             },
             {
               "required": ["contents"]
-            },
-            {
-              "required": ["guidance"]
             }
           ]
         }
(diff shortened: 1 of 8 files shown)


petechd:
[Banner on introduction page (3 SP)](https://trello.com/c/Dog3Cv1U/5169-banner-on-introduction-page-3-sp)

jamesnunn (approved):

schemas/common_definitions.json
```diff
@@ -51,6 +54,15 @@
         "description": {
           "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/string_with_placeholders"
         },
+        "guidance": {
```
MebinAbraham, line 57: Content blocks are for generic content. The guidance should be a controlled structure. It won't make sense for us to allow guidance for all content blocks. This makes sense as part of primary content in introduction. Same way we have allowed guidance for question.

MebinAbraham (commented):
Worth adding example to the valid schema also,

schemas/common_definitions.json
```diff
@@ -406,5 +406,14 @@
...
+  },
+  "introduction_guidance": {
+    "description": "Introduction guidance",
+    "type": "object",
+    "properties": {
+      "contents": {
+        "$ref": "https://eq.ons.gov.uk/common_definitions.json#/contents"
+      }
+    }
   }
```
MebinAbraham, line 418: We should name this just "guidance", then use it for both intro and question page. Atm, we have another `question_guidance` which is a duplicate.

schemas/blocks/introduction.json
```diff
@@ -30,9 +30,13 @@
...
             "contents": {
               "$ref": "https://eq.ons.gov.uk/common_definitions.json#/contents"
+            },
+            "guidance": {
+              "$ref": "https://eq.ons.gov.uk/common_definitions.json#/introduction_guidance"
             }
           },
-          "additionalProperties": false
+          "additionalProperties": true,
+          "required": ["id", "title"]
```
MebinAbraham, line 39: Is title required?. What does runner expect?. Most likely one of, title, guidance or content?.

tests/schemas/valid/test_introduction_with_guidance.json
```diff
@@ -0,0 +1,258 @@
+{
```
MebinAbraham, line 1: Not a blocker but a lot of content in this can be slimmed down. Most of the content isn't needed, placeholders etc.

schemas/common_definitions.json
petechd: Ok, moved it to `primary_content` so nesting removed and guidance not part of generic content anymore.

tests/schemas/valid/test_introduction_with_guidance.json
petechd: I will reduce it if time allows.

schemas/blocks/introduction.json
petechd: `id` required, the rest part of `anyOf` as discussed. Thanks.

schemas/common_definitions.json
petechd: So only `guidance` now, thanks!

MebinAbraham (approved):

PR #144, opened 2022-05-23
Add placeholder transform unit and percentage definitions

petechd:
### PR Context
This introduces new transform definitions to support unit and percentage placeholder transforms added in runner.

### Checklist


petechd:
[Placeholder to support piping of unit and percentage answers (3 SP)](https://trello.com/c/OZVLqMQ3/5286-placeholder-to-support-piping-of-unit-and-percentage-answers-3-sp)

schemas/string_interpolation/transforms/format_unit.json
```diff
@@ -0,0 +1,31 @@
...
+      "transform": {
+        "type": "string",
+        "enum": ["format_unit"]
+      },
+      "arguments": {
+        "type": "object",
+        "properties": {
+          "value": {
+            "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/placeholder_value_sources"
+          },
```
MebinAbraham, line 16: I think we should add some validation to ensure if an answer value source is used, then that answer's type matches the `unit`, otherwise we should raise a validation error.

schemas/string_interpolation/transforms/format_percentage.json
```diff
@@ -0,0 +1,25 @@
...
+      "transform": {
+        "type": "string",
+        "enum": ["format_percentage"]
+      },
+      "arguments": {
+        "type": "object",
+        "properties": {
+          "value": {
+            "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/placeholder_value_sources"
+          }
```
MebinAbraham, line 16: Same as the one below, I think it's worth validating, if an answer source is used, that it is in fact a percentage answer type.

schemas/string_interpolation/transforms/format_unit.json
```diff
@@ -0,0 +1,31 @@
...
+          "length": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
+          },
+          "unit": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
+          }
+        },
+        "additionalProperties": false,
+        "required": ["value", "unit"]
```
MebinAbraham, line 25: For length, a description should be added. And we should also use `default` to denote what we fall back to.

schemas/string_interpolation/transforms/format_unit.json
```diff
@@ -0,0 +1,31 @@
...
+      },
+      "arguments": {
+        "type": "object",
+        "properties": {
+          "value": {
+            "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/placeholder_value_sources"
+          },
+          "length": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
+          },
```
MebinAbraham, line 19: This should be an enum as it can only be either `short` or `long` or `narrow`. I think also `unit_length` would be clearer and would match what it is called in answer objects.
```suggestion
          "unit_length": {
            "type": "string",
            "enum": ["short", "narrow", "long"]
          },
```

schemas/string_interpolation/transforms/format_unit.json
```diff
@@ -0,0 +1,31 @@
...
+          "unit": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
+          }
```
MebinAbraham, line 22: This shouldn't be any arbitrary string either. It is a set of predefined names. We have a list already.

```suggestion
          "unit": {
            "$ref": "https://eq.ons.gov.uk/lists/units.json#/unit"
          }
```

schemas/string_interpolation/definitions.json
```diff
@@ -76,6 +76,12 @@
             {
               "$ref": "https://eq.ons.gov.uk/string_interpolation/transforms/format_number.json#/format_number"
             },
+            {
+              "$ref": "https://eq.ons.gov.uk/string_interpolation/transforms/format_percentage.json#/format_percentage"
+            },
+            {
+              "$ref": "https://eq.ons.gov.uk/string_interpolation/transforms/format_unit.json#/format_unit"
+            },
```
MebinAbraham, line 84: We should add these to a valid schema also.

petechd: Added a valid schema from runner, with format unit and percentage transforms.

schemas/string_interpolation/transforms/format_unit.json
petechd: Ok, so I added new validation to validate exactly this.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,6 +108,19 @@ def validate_option_label_from_value_placeholder(self, answer_id):
                 identifier=answer_id,
             )
 
+    def validate_answer_type_correct(self, argument, transform_type):
+        if argument.get("source") == "answers":
+            transform_type = transform_type.split("_")[1]
```
MebinAbraham, line 113: I personally prefer it to be explicit than coupling with our naming scheme, but not a blocker.

app/error_messages.py
```diff
@@ -5,6 +5,7 @@
     "The referenced answer cannot be used to set the minimum of answer"
 )
 ANSWER_TYPE_FOR_OPTION_LABEL_FROM_VALUE_INVALID = "The answer type for option label from value is not of type ['Radio','Checkbox','Dropdown']"
+ANSWER_TYPE_FOR_TRANSFORM_TYPE_INVALID = "The answer type for 'format_percentage' or 'format_unit' transform is not of type 'Percentage','Unit'"
```
MebinAbraham, line 8: This should be specific. We want to know if it was one or the other etc. I would make this a str like:

```py
ANSWER_TYPE_FOR_TRANSFORM_TYPE_INVALID = "The answer type for '{transform}' transform is not of type '{answer_type}'"
```

Then you can just use `.format()` when calling this to pass the relevant info.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,6 +108,19 @@ def validate_option_label_from_value_placeholder(self, answer_id):
                 identifier=answer_id,
             )
 
+    def validate_answer_type_correct(self, argument, transform_type):
```
MebinAbraham, line 111: The naming is too vague. We should update to something like `validate_answer_type_for_transform`

app/validators/placeholders/placeholder_validator.py
```diff
@@ -123,6 +136,13 @@ def validate_placeholder_transforms(self, transforms):
                 and argument_name == "answer_id"
             ):
                 self.validate_option_label_from_value_placeholder(argument)
+            if (
+                first_transform["transform"] in ["format_unit", "format_percentage"]
+                and argument_name == "value"
+            ):
+                self.validate_answer_type_correct(
+                    argument, first_transform["transform"]
+                )
```
MebinAbraham, line 145: This needs to be done for all transforms. This is only validating the first one.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,6 +108,19 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
             )
 
+    def validate_answer_type_correct(self, argument, transform_type):
+        if argument.get("source") == "answers":
+            transform_type = transform_type.split("_")[1]
+            answer_id = argument.get("identifier")
+            answer_type = self.questionnaire_schema.answers_with_context[answer_id][
+                "answer"
+            ]["type"]
+            if answer_type.lower() != transform_type:
```
MebinAbraham, line 118: For units, we need to also validate the `unit` itself matches. The answer cant set unit to `length-meter` and then have transformed state `length-mile`, they should match.

tests/schemas/valid/test_placeholder_default_value.json
```diff
@@ -0,0 +1,284 @@
+{
```
MebinAbraham, line 1: I don't think we need an entire new schema. Can we not add two new questions into an existing one?

schemas/string_interpolation/transforms/format_unit.json
```diff
@@ -0,0 +1,34 @@
...
+        "enum": ["format_unit"]
+      },
+      "arguments": {
+        "type": "object",
+        "properties": {
+          "value": {
+            "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/placeholder_value_sources"
+          },
+          "unit_length": {
+            "description": "A babel units parameter, possible values are “short”, “long” or “narrow”",
```
MebinAbraham, line 18:
```suggestion
            "description": "The formatting length. e.g. 1h (narrow) / 1 hr (short) / 1 hour (long)",
```

app/validators/placeholders/placeholder_validator.py
petechd: But how do I do that since we pass two types of transforms into here?

app/error_messages.py
petechd: Ok, added and using `.format()`.

app/validators/placeholders/placeholder_validator.py
petechd: Renamed, thanks!

app/validators/placeholders/placeholder_validator.py
petechd: Ok, added thanks!

app/validators/placeholders/placeholder_validator.py
petechd: Ok, added extra method of validation to check for unit mismatches.

tests/schemas/valid/test_placeholder_default_value.json
petechd: Ok added percentage and unit transforms showcase questions and interstitials to existing placeholder transforms schema, thanks!

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,6 +108,46 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
 
+    def validate_answer_type_for_transform(
+        self, argument, argument_name, transform_type
+    ):
+        if (
+            transform_type in ["format_unit", "format_percentage"]
+            and argument_name == "value"
+            and argument.get("source") == "answers"
+        ):
+            answer_id = argument.get("identifier")
```
MebinAbraham, line 119: Can be key access. As at this point, this has to exist.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,6 +108,46 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
+            answer_type = self.questionnaire_schema.answers_with_context[answer_id][
+                "answer"
+            ]["type"]
+            if answer_type.lower() not in transform_type:
+                self.add_error(
+                    error_messages.ANSWER_TYPE_FOR_TRANSFORM_TYPE_INVALID.format(
+                        transform=transform_type, answer_type=answer_type
+                    ),
+                    identifier=answer_id,
+                )
```
MebinAbraham, line 129: Minor, can we use guard clauses to escape early to avoid the nesting?

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,6 +108,46 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
+    def validate_answer_and_transform_unit_match(self, arguments, transform_type):
+        if transform_type == "format_unit":
+            answer_id = arguments["value"].get("identifier")
+            unit = arguments["unit"]
+            if (
+                unit
+                != self.questionnaire_schema.answers_with_context[answer_id]["answer"][
+                    "unit"
+                ]
+            ):
```
MebinAbraham, line 140: Minor, can we use guard clauses to escape early to avoid the nesting?

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,6 +108,46 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
+                    error_messages.ANSWER_UNIT_AND_TRANSFORM_UNIT_MISMATCH.format(
+                        answer_unit=self.questionnaire_schema.answers_with_context[
+                            answer_id
+                        ]["answer"]["unit"],
+                        transform_unit=unit,
+                    ),
+                    identifier=answer_id,
+                )
+
     def validate_placeholder_transforms(self, transforms):
```
MebinAbraham, line 151: I think we need to clean up this method, there are duplicate validation calls which we should avoid. We should look at doing something like:


```py
    def _validate_placeholder_previous_transforms(self, transforms):
        # First transform can't reference a previous transform
        first_transform = transforms[0]
        for argument_name in first_transform.get("arguments"):
            argument = first_transform["arguments"][argument_name]
            if (
                isinstance(argument, dict)
                and argument.get("source") == "previous_transform"
            ):
                self.add_error(self.FIRST_TRANSFORM_CONTAINS_PREVIOUS_TRANSFORM_REF)

        # Previous transform must be referenced in all subsequent transforms
        for transform in transforms[1:]:
            previous_transform_used = False
            for argument_name in transform.get("arguments"):
                argument = transform["arguments"][argument_name]
                if (
                    isinstance(argument, dict)
                    and argument.get("source") == "previous_transform"
                ):
                    previous_transform_used = True

            if not previous_transform_used:
                self.add_error(self.NO_PREVIOUS_TRANSFORM_REF_IN_CHAIN)

    def validate_placeholder_transforms(self, transforms):
        self._validate_placeholder_previous_transforms(transforms)

        for transform in transforms:
            for argument_name in transform.get("arguments"):
                argument = transform["arguments"][argument_name]
                if (
                    transform["transform"] == "option_label_from_value"
                    and argument_name == "answer_id"
                ):
                    self.validate_option_label_from_value_placeholder(argument)

                self.validate_answer_type_for_transform(
                    transform.get("arguments"), argument_name, transform["transform"]
                )

            self.validate_answer_and_transform_unit_match(
                transform.get("arguments"), transform["transform"]
            )
```

The current methods could do some work to make it consistent with what we have, existing check transform before calling, and the new ones call and then check the transform. However, won't block on this as it is minor.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,6 +108,46 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
+            ]["type"]
+            if answer_type.lower() not in transform_type:
+                self.add_error(
+                    error_messages.ANSWER_TYPE_FOR_TRANSFORM_TYPE_INVALID.format(
+                        transform=transform_type, answer_type=answer_type
+                    ),
+                    identifier=answer_id,
+                )
+
+    def validate_answer_and_transform_unit_match(self, arguments, transform_type):
```
MebinAbraham, line 131: I know we haven't done it for all, but given we dont have type checks for this repo, worth enforcing keyword arguments for all these methods to avoid passing args in the wrong order. So using `*, ` etc.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,6 +108,46 @@ def validate_option_label_from_value_placeholder(self, answer_id):
                 identifier=answer_id,
             )
 
+    def validate_answer_type_for_transform(
```
MebinAbraham, line 111: All the new changes needs to be unit tested.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -111,42 +111,46 @@ def validate_option_label_from_value_placeholder(self, answer_id):
     def validate_answer_type_for_transform(
         self, argument, argument_name, transform_type
     ):
-        if (
+        if not (
             transform_type in ["format_unit", "format_percentage"]
             and argument_name == "value"
             and argument.get("source") == "answers"
         ):
```
MebinAbraham, line 118: Double negative should be avoided.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -111,42 +111,46 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
+        if transform_type != "format_unit":
+            return
+        answer_id = arguments["value"].get("identifier")
+        unit = arguments["unit"]
+        if not (
+            unit
+            != self.questionnaire_schema.answers_with_context[answer_id]["answer"][
+                "unit"
+            ]
+        ):
```
MebinAbraham, line 143: Double negative should be avoided. Use `==` instead of `not !=`

app/validators/placeholders/placeholder_validator.py
```diff
@@ -111,42 +111,46 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
-                    error_messages.ANSWER_UNIT_AND_TRANSFORM_UNIT_MISMATCH.format(
-                        answer_unit=self.questionnaire_schema.answers_with_context[
-                            answer_id
-                        ]["answer"]["unit"],
-                        transform_unit=unit,
-                    ),
-                    identifier=answer_id,
-                )
+        if transform_type != "format_unit":
+            return
```
MebinAbraham, line 135: I prefer to `return None` so that it is explicit. We have enforced this in runner. Not enforced here but worth it for consistency.

app/validators/placeholders/placeholder_validator.py
petechd: Done, thanks!

app/validators/placeholders/placeholder_validator.py
petechd: Done, thanks!

app/validators/placeholders/placeholder_validator.py
petechd: Added, great suggestion, thanks!

app/validators/placeholders/placeholder_validator.py
petechd: Done, thanks!

app/validators/placeholders/placeholder_validator.py
petechd: Done, thanks!

app/validators/placeholders/placeholder_validator.py
petechd: Using "any" here now, is that better? Was there double negation here in the first place?

app/validators/placeholders/placeholder_validator.py
petechd: Added keyword arguments when invoking, I hope that is what you meant?

app/validators/placeholders/placeholder_validator.py
petechd: Added unit tests to test changes.

rmccar (approved):

app/validators/placeholders/placeholder_validator.py
MebinAbraham: Done seem to be done.

tests/schemas/invalid/test_invalid_placeholder_answer_type_from_transform.json
```diff
@@ -0,0 +1,103 @@
+{
+  "mime_type": "application/json/ons/eq",
+  "language": "en",
+  "schema_version": "0.0.1",
+  "data_version": "0.0.3",
+  "survey_id": "0",
+  "title": "Placeholder Test",
+  "theme": "default",
+  "description": "A questionnaire to test placeholders",
```
MebinAbraham, line 9: This should explain why this is an invalid schema.

app/validators/placeholders/placeholder_validator.py
petechd: Ok added `*,` so only keyword args accepted, thanks for the read!

app/validators/placeholders/placeholder_validator.py
petechd: For some reason I haven't pushed it, don't know what happened. Now it is there, thanks!

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,6 +108,50 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
+        answer_id = arguments["value"].get("identifier")
+        unit = arguments["unit"]
+        if (
+            unit
+            == self.questionnaire_schema.answers_with_context[answer_id]["answer"][
+                "unit"
+            ]
+        ):
+            return None
+        self.add_error(
```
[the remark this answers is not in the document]
petechd: Added, thanks!

tests/schemas/invalid/test_invalid_placeholder_answer_type_from_transform.json
petechd: Added explanation thanks!

petechd:
All changes done (including that outstanding refactor), had to comment here otherwise my comment  about addressing all the change requests are hard to spot.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,7 +108,53 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
+            error_messages.ANSWER_TYPE_FOR_TRANSFORM_TYPE_INVALID.format(
+                transform=transform_type, answer_type=answer_type
+            ),
+            identifier=answer_id,
+        )
+
+    def validate_answer_and_transform_unit_match(self, *, arguments, transform_type):
+        if transform_type != "format_unit":
+            return None
+        answer_id = arguments["value"].get("identifier")
```
MebinAbraham, line 136: Minor but looks squashed.

MebinAbraham (commented):
Just minor issues

app/validators/placeholders/placeholder_validator.py
petechd: Did this:
```python
value = arguments["value"]
        answer_id = value.get("identifier")
        unit = arguments["unit"]
```
plus added some spaces.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,7 +108,53 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
+            and argument.get("source") == "answers"
+        ):
+            return None
+        answer_id = argument["identifier"]
+        answer_type = self.questionnaire_schema.answers_with_context[answer_id][
+            "answer"
+        ]["type"]
+        if answer_type.lower() in transform_type:
+            return None
+        self.add_error(
```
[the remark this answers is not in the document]
petechd: Added some new line separators, thanks.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,7 +108,57 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
-    def validate_placeholder_transforms(self, transforms):
+    def validate_answer_type_for_transform(
+        self, argument, argument_name, transform_type
+    ):
+        if not (
+            transform_type in ["format_unit", "format_percentage"]
+            and argument_name == "value"
+            and argument.get("source") == "answers"
+        ):
+            return None
```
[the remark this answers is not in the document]
petechd: Seems like it fails validation in the right place, I've added invalid test schema with `Number` type answer to transform and an extra test to showcase the behaviour and we get the right validation error message (`The answer type for 'format_percentage' transform is not of type 'Number'`).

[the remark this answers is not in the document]
petechd: Ok after our discussion I finally get it, nesting is corrected now, had to change the `if` logic slightly, thanks!

app/validators/placeholders/placeholder_validator.py
```diff
@@ -114,24 +114,25 @@ def validate_answer_type_for_transform(
...
-            and argument.get("source") == "answers"
+            and argument.get("value")
         ):
             return None
 
-        answer_id = argument["identifier"]
-        answer_type = self.questionnaire_schema.answers_with_context[answer_id][
-            "answer"
-        ]["type"]
+        if argument["value"].get("source") and argument["value"]["source"] == "answers":
```
MebinAbraham, line 121: This can be checked in the statement above. We don't need 2 seperate ifs.

petechd: Ok, added one statement that checks all, thanks!

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,7 +108,58 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
+        value = arguments["value"]
+        answer_id = value.get("identifier")
+        unit = arguments["unit"]
+
+        if (
+            unit
+            == self.questionnaire_schema.answers_with_context[answer_id]["answer"][
+                "unit"
+            ]
+        ):
```
MebinAbraham, line 149: This currently raises a key error if the answer_id provided is for a non-unit answer.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,7 +108,58 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
+
+        if answer_type.lower() in transform_type:
+            return None
+
+        self.add_error(
+            error_messages.ANSWER_TYPE_FOR_TRANSFORM_TYPE_INVALID.format(
+                transform=transform_type, answer_type=answer_type
+            ),
+            identifier=answer_id,
+        )
```
MebinAbraham, line 134: The error message is not quite right. Currently, you get:
```
"The answer type for 'format_percentage' transform is not of type 'Number'"
```
Should be `"The answer type for 'format_percentage' transform is not of type 'Percentage'"`.

Maybe we should make it more meaningful, so something like:
`"Expected the answer type for 'format_percentage' transform to be type 'Percentage'" but got type 'Number'`.

petechd: Added the second message as it's much more meaningful.

app/validators/placeholders/placeholder_validator.py
petechd: Added condition here so the method is called only when no errors yet👍.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -108,7 +108,61 @@ def validate_option_label_from_value_placeholder(self, answer_id):
...
+        value = arguments["value"]
+        answer_id = value.get("identifier")
+        unit = arguments["unit"]
+
+        if not self.errors and (
+            unit
+            == self.questionnaire_schema.answers_with_context[answer_id]["answer"][
+                "unit"
+            ]
+        ):
```
MebinAbraham, line 152: This should be `if error or unit_match: return None`. This would check it if there was error. We want to exit early if there are errors or if there isnt error but unit does match.

petechd: Added thanks!

MebinAbraham (approved):
👍

PR #145, opened 2022-05-26
Validate calculated summary source types in when rules

berroar:
### PR Context

Adds support for the validation of calculated summary source types in the new when rules.

Added a valid example schema `tests/schemas/valid/test_routing_calculated_summary_dependencies.json`.

The change is required to support this runner [PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/868)

### Checklist


berroar:
[Check Calculated Summary source answer type in new when rules](https://trello.com/c/vhH5StJl/5319-check-calculated-summary-source-answer-type-in-new-when-rules)

app/validators/routing/types.py
```diff
@@ -43,13 +43,22 @@
...
+        answer_type = schema.answers_with_context[answer_id]["answer"]["type"]
         return ANSWER_TYPE_TO_JSON_TYPE[answer_type]
 
+    if source == "calculated_summary":
+        answer_type_list = []
+        block_id = schema.get_block(value_source["identifier"])
+        answer_ids = block_id["calculation"]["answers_to_calculate"]
+        for answer_id in answer_ids:
+            answer_type = schema.answers_with_context[answer_id]["answer"]["type"]
+            answer_type_list.append(ANSWER_TYPE_TO_JSON_TYPE[answer_type])
```
MebinAbraham, line 59: We only need to check 1 answer id. We should have other validation to ensure answers within calculated summary are of all same type (existing gap, so separate card).

MebinAbraham, line 59: We only need to check 1 answer id. We should have other validation to ensure answers within calculated summary are of all same type (existing gap, so separate card).

berroar: Updated this 👍

berroar:
Delted duplicate comment

app/validators/routing/types.py
petechd: to avoid using `0` index of the list we can potentially do something like:
```python   if source == "calculated_summary":
        block_id = schema.get_block(value_source["identifier"])
        for answer in block_id["calculation"]["answers_to_calculate"].keys():
            answer_id = block_id["calculation"]["answers_to_calculate"][answer]
            answer_type = schema.answers_with_context[answer_id]["answer"]["type"]
```
so you only go through the loop once (and return early).

berroar: Updated to use `next(iter(block_id["calculation"]["answers_to_calculate"]))` as discussed 👍

petechd (approved):

MebinAbraham (approved):

app/validators/routing/types.py
```diff
@@ -43,11 +43,17 @@
...
     source = value_source["source"]
     if source == "answers":
         answer_id = value_source["identifier"]
-        answer_type = answers_with_context[answer_id]["answer"]["type"]
+        answer_type = schema.answers_with_context[answer_id]["answer"]["type"]
+        return ANSWER_TYPE_TO_JSON_TYPE[answer_type]
+
+    if source == "calculated_summary":
+        block_id = schema.get_block(value_source["identifier"])
+        answer_id = next(iter(block_id["calculation"]["answers_to_calculate"]))
```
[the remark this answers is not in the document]
berroar: Reverted this

PR #148, opened 2022-06-30
Allow radios as mutually exclusive answers

rmccar:
### PR Context
This PR will allow radios as acceptable answers for mutually exclusive options

It will also fix an issue with the docs in the readme and 2 typos in a variable names

### Checklist

### DIFF
--- a/app/validators/questions/mutually_exclusive_validator.py
+++ b/app/validators/questions/mutually_exclusive_validator.py
@@ -16,7 +16,7 @@ class MutuallyExclusiveQuestionValidator(QuestionValidator):
 
         if (
             self.answers[-1]["type"] != "Checkbox"
-            or self.answers[-1]["type"] != "Radio"
+            and self.answers[-1]["type"] != "Radio"
         ):
             self.add_error(
                 self.NON_CHECKBOX_RADIO_ANSWER, answer_id=self.answers[-1]["id"]
--- a/app/validators/questions/mutually_exclusive_validator.py
+++ b/app/validators/questions/mutually_exclusive_validator.py
@@ -1,6 +1,7 @@
+from app.answer_type import AnswerType
 from app.validators.questions.question_validator import QuestionValidator
 
-from app.answer_type import AnswerType
+
 class MutuallyExclusiveQuestionValidator(QuestionValidator):
     question = {}
     MUTUALLY_EXCLUSIVE_CONTAINS_MANDATORY = (
@@ -14,7 +15,10 @@ class MutuallyExclusiveQuestionValidator(QuestionValidator):
         if any(answer["mandatory"] is True for answer in self.answers):
             self.add_error(self.MUTUALLY_EXCLUSIVE_CONTAINS_MANDATORY)
 
-        if (AnswerType(self.answers[-1]["type"]) not in {AnswerType.CHECKBOX, AnswerType.RADIO}):
+        if AnswerType(self.answers[-1]["type"]) not in {
+            AnswerType.CHECKBOX,
+            AnswerType.RADIO,
+        }:
             self.add_error(
                 self.NON_CHECKBOX_RADIO_ANSWER, answer_id=self.answers[-1]["id"]
             )
--- a/app/validators/questions/mutually_exclusive_validator.py
+++ b/app/validators/questions/mutually_exclusive_validator.py
@@ -7,7 +7,7 @@ class MutuallyExclusiveQuestionValidator(QuestionValidator):
     MUTUALLY_EXCLUSIVE_CONTAINS_MANDATORY = (
         "MutuallyExclusive question type cannot contain mandatory answers."
     )
-    NON_CHECKBOX_RADIO_ANSWER = "Question is not of type Checkbox or Radio."
+    INVALID_EXCLUSIVE_ANSWER = "Mutually exclusive answer is not of type Checkbox or Radio."
     NON_EXCLUSIVE_RADIO_ANSWER = (
         "Mutually exclusive questions cannot contain non exclusive Radio answers."
     )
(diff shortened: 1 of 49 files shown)


rmccar:
[Multiple mutually exclusive options (5 SP)](https://trello.com/c/IVa0CArU/5116-multiple-mutually-exclusive-options-5-sp)

app/validators/questions/mutually_exclusive_validator.py
```diff
@@ -6,14 +6,19 @@ class MutuallyExclusiveQuestionValidator(QuestionValidator):
...
 
         if any(answer["mandatory"] is True for answer in self.answers):
             self.add_error(self.MUTUALLY_EXCLUSIVE_CONTAINS_MANDATORY)
 
-        if self.answers[-1]["type"] != "Checkbox":
-            self.add_error(self.NON_CHECKBOX_ANSWER, answer_id=self.answers[-1]["id"])
+        if (
+            self.answers[-1]["type"] != "Checkbox"
+            and self.answers[-1]["type"] != "Radio"
+        ):
```
MebinAbraham, line 20: We should use the `AnswerType` enum here and also this should be replaced by an `in` check. so `type not in [checkbox, radio]`

app/validators/questions/mutually_exclusive_validator.py
```diff
@@ -6,14 +6,19 @@ class MutuallyExclusiveQuestionValidator(QuestionValidator):
...
         super().validate()
 
         if any(answer["mandatory"] is True for answer in self.answers):
             self.add_error(self.MUTUALLY_EXCLUSIVE_CONTAINS_MANDATORY)
 
-        if self.answers[-1]["type"] != "Checkbox":
-            self.add_error(self.NON_CHECKBOX_ANSWER, answer_id=self.answers[-1]["id"])
+        if (
+            self.answers[-1]["type"] != "Checkbox"
+            and self.answers[-1]["type"] != "Radio"
```
MebinAbraham, line 19: Now that we allow radios. We should validate that the other answers isn't a Radio. I would loop though all answers that isn't the last one, as soon we will have multiple answers that aren't the exclusive answer.

MebinAbraham (commented):
A new valid schema needs adding, and the code changes need unit testing.

app/validators/questions/mutually_exclusive_validator.py
rmccar: > Now that we allow radios. We should validate that the other answers isn't a Radio. I would loop though all answers that isn't the last one, as soon we will have multiple answers that aren't the exclusive answer.

Is this needed? Because the DS doesn't allow radios to have mutually exclusive options

rmccar:
> A new valid schema needs adding, and the code changes need unit testing.

I think the current tests cover the logic because they test if the mutually exclusive option is not a radio or checkbox but I have added a valid schema question that tests if the mutually exclusive option is a radio

app/validators/questions/mutually_exclusive_validator.py
MebinAbraham: Validator is stop invalid schemas. Because the DS doesn't allow it, is the reason why we need to validate, otherwise it would error at run time.

app/validators/questions/mutually_exclusive_validator.py
```diff
@@ -6,14 +7,28 @@ class MutuallyExclusiveQuestionValidator(QuestionValidator):
     MUTUALLY_EXCLUSIVE_CONTAINS_MANDATORY = (
         "MutuallyExclusive question type cannot contain mandatory answers."
     )
-    NON_CHECKBOX_ANSWER = "Question is not of type Checkbox."
+    NON_CHECKBOX_RADIO_ANSWER = "Question is not of type Checkbox or Radio."
```
MebinAbraham, line 10: This might be better as `INVALID_EXCLUSIVE_ANSWER`. I think the error text should mention mutually exclusive as it doesn't make sense without it.

tests/schemas/valid/test_question_validation_messages.json
```diff
@@ -186,6 +186,56 @@
...
+                      }
+                    ]
+                  }
+                ],
+                "validation": {
+                  "messages": {
+                    "MANDATORY_QUESTION": "Select an answer to continue"
+                  }
+                }
+              }
```
MebinAbraham, line 238: Minor, but minimise schema to the bare minimum. These are required.

app/validators/questions/mutually_exclusive_validator.py
```diff
@@ -6,14 +7,28 @@ class MutuallyExclusiveQuestionValidator(QuestionValidator):
...
+        if AnswerType(self.answers[-1]["type"]) not in {
+            AnswerType.CHECKBOX,
+            AnswerType.RADIO,
+        }:
+            self.add_error(
+                self.NON_CHECKBOX_RADIO_ANSWER, answer_id=self.answers[-1]["id"]
+            )
+
+        if any(
+            (AnswerType(answer["type"]) == AnswerType.RADIO) is True
```
MebinAbraham, line 30: The `is True` bit is not needed, it always evaluates to a bool.

app/validators/questions/mutually_exclusive_validator.py
```diff
@@ -6,14 +7,28 @@ class MutuallyExclusiveQuestionValidator(QuestionValidator):
...
+            for answer in self.answers[:-1]
+        ):
+            self.add_error(self.NON_EXCLUSIVE_RADIO_ANSWER)
```
MebinAbraham, line 33: This doesn't seem to be unit tested.

MebinAbraham (commented):
Minor issues.

tests/schemas/valid/test_mutually_exclusive.json
```diff
@@ -0,0 +1,95 @@
...
+                        "value": "I dont have a house"
+                      }
+                    ]
+                  }
+                ],
+                "validation": {
+                  "messages": {
+                    "MANDATORY_QUESTION": "Select an answer to continue"
+                  }
+                }
```
MebinAbraham, line 87: Not needed.

MebinAbraham (approved):

tests/schemas/valid/test_mutually_exclusive.json
petechd: Removed!

berroar (approved):

PR #150, opened 2022-07-20
Add fix to calculations value source validation

petechd:
### PR Context
Currently our calculations value source validation breaks when value for calculation is not a dict type, this fixes the issue. Valid schema added with previously breaking `value` for calculations.

### Checklist

### DIFF
--- a/tests/schemas/valid/test_calculations_value_source.json
+++ b/tests/schemas/valid/test_calculations_value_source.json
@@ -71,10 +71,7 @@
                   {
                     "calculation_type": "sum",
                     "value": 100,
-                    "answers_to_calculate": [
-                      "breakdown-1",
-                      "breakdown-2"
-                    ],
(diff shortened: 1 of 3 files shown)


berroar (approved):

rmccar (approved):

tests/schemas/valid/test_calculations_value_source.json
```diff
@@ -0,0 +1,81 @@
+{
```
MebinAbraham, line 1: The schema name doesn't match the content. This schema has not value source...?

petechd: I meant value as general "source" for calculations here, renamed to `test_calculations_with_numeric_value.json` or should it be shorter: `test_calculations_numeric_value.json`?

MebinAbraham (approved):

PR #155, opened 2022-09-30
Upgrade to Python 3.10

Yuyuutsu:
### PR Context
This PR is upgrading the python version from 3.9 to 3.10 to ensure we have the latest features. Dockerfile and Poetry .toml file also updated.

### How to review
Check if tests and Makefile scripts are working fine. Run local copy of validator against local schemas.

### Checklist


Yuyuutsu:
[Upgrade non-runner Python repos to 3.10 (2 SP)](https://trello.com/c/B9pw2uPW/5229-upgrade-non-runner-python-repos-to-310-2-sp)

petechd (approved):
All the local scripts, tests and api works fine with 3.10.7 and new version of poetry.

pyproject.toml
```diff
@@ -8,7 +8,7 @@ description = "API for questionnaire schema validation"
 authors = ["ONSDigital"]
```
MebinAbraham, line 8: I suggest unpinning `black` dev package.

pyproject.toml
```diff
@@ -21,12 +21,12 @@ eq-translations = {git = "https://github.com/ONSDigital/eq-translations.git"}
...
 pytest = "^6.2.5"
 flake8 = "^4.0.1"
-pylint = "^2.12.2"
+pylint = "^2.15.3"
 pylint-quotes = "^0.2.3"
 pytest-cov = "^3.0.0"
 pytest-xdist = "^2.5.0"
 pytest-sugar = "^0.9.4"
-black = "21.10b0"
+black = "^21.10b0"
```
petechd, line 29: You might have to pin it to `"^22.10.0"` here since the previous pin will only update up to latest `21` version.

Yuyuutsu: updated, thanks !

pyproject.toml
```diff
@@ -21,12 +21,12 @@ eq-translations = {git = "https://github.com/ONSDigital/eq-translations.git"}
 [tool.poetry.dev-dependencies]
 pytest = "^6.2.5"
 flake8 = "^4.0.1"
-pylint = "^2.12.2"
+pylint = "^2.15.3"
```
petechd, line 24: There is a later version of pylint so maybe if you pin to that it will solve your linting issue (`"^2.15.4"`).

Yuyuutsu: Cyclic import still occurs. I cannot create the error locally, only shows up during PR

.pylintrc
```diff
@@ -1,420 +1,621 @@
...
+# --disable=W".
 disable=
     too-many-arguments,
     too-many-instance-attributes,
     too-few-public-methods,
     missing-docstring,
     invalid-name,
     inconsistent-return-statements,
-    bad-continuation,
+    cyclic-import
```
berroar, line 152: I don't think we should be ignoring all of these, I think we can possibly fix the issues pylint highlighted on your failed runs. Happy to have a look with you if it is causing issues

Yuyuutsu: Yeah that sounds good 👍

Yuyuutsu: Fixed

berroar (approved):

Yuyuutsu:
Recent pushes was signing recent commits

PR #156, opened 2022-10-25
Validate theme in JSON schema

Yuyuutsu:
### PR Context
The theme property in schemas can be set to any string which can cause the survey fail to launch. This PR validates that the theme used is the ones supported by the eQ Runner.

### Checklist
-  Passes `make validate-test-schemas ` in eQ Runner
- * [ ] eq-translations updated to support any new schema keys which need translation

### DIFF
--- a/schemas/questionnaire_v1.json
+++ b/schemas/questionnaire_v1.json
@@ -46,7 +46,8 @@
       "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
     },
     "theme": {
-      "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
+      "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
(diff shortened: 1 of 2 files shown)


Yuyuutsu:
[Validate theme in JSON schema (1 SP)](https://trello.com/c/OCYE5LSP/5412-validate-theme-in-json-schema-1-sp)

petechd (commented):
We should add an invalid schema in `tests/schemas/invalid` (with incorrect theme) to test it.

schemas/questionnaire_v1.json
```diff
@@ -46,7 +46,16 @@
       "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
     },
     "theme": {
-      "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string"
+      "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
```
petechd, line 49: As we discussed, this can't be both non empty string and enum.

MebinAbraham (approved):

petechd (approved):

PR #157, opened 2022-11-07
Re-enable q_codes validation

petechd:
### PR Context
We can re-enable our existing validation since `q_codes` gaps in every business schema will be fixed soon.

### Checklist
3 unit tests are checking if test schema is validated correctly.


petechd:
[Enable q-codes validation (1 SP)](https://trello.com/c/LcUtQAWV/5457-enable-q-codes-validation-1-sp)

berroar (approved):

rmccar (approved):

PR #158, opened 2022-11-30
Add support for new calculated summary schema structure

berroar:
### PR Context

Adds support for the new CalculatedSummary schemas structure:

Updated format:
```
"calculation": {
   "operation": {
       "+": [
           {
               "source": "answers",
               "identifier": "first-number-answer"
           },
           {
               "source": "answers",
               "identifier": "second-number-answer"
           },
           {
               "source": "answers",
               "identifier": "third-number-answer"
           }
       ]
   },
   "title": "Grand total of previous values"
},
```

Adds additional validation checks on CalcualtedSummary blocks, to ensure that:
  - Answer ids referenced in CalculatedSummary blocks are set before the CalculatedSummary block.
  - Answer ids used by the CalculatedSummary are in the same section as the CalculatedSummary block.

### Checklist


berroar:
[Refactor calculated summary (5 SP)](https://trello.com/c/qVCAwIZQ/5350-refactor-calculated-summary-5-sp)

rmccar (approved):

app/validators/blocks/calculated_summary_block_validator.py
```diff
@@ -49,3 +57,35 @@ def validate(self):
...
+                self.ANSWER_SET_IN_DIFFERENT_SECTION_FOR_CALCULATED_SUMMARY,
+                block=self.block,
+            )
+
+    def get_calculated_answer_ids(self):
+        if self.block["calculation"].get("answers_to_calculate"):
+            return self.block["calculation"]["answers_to_calculate"]
+        return [
+            value["identifier"]
+            for value in self.block["calculation"]["operation"]["+"]
```
MebinAbraham, line 89: We shouldn't expect it to always be one level. `+` will allow recursion, so this should get all identifier at all levels. I think we must have a function to return all values for identifiers for a given schema such as `self.block["calculation"]["operation"]`

app/validators/questionnaire_schema.py
```diff
@@ -412,9 +412,18 @@ def _get_numeric_value_for_value_source(
...
+                    "answers_to_calculate"
+                ]
+            else:
+                answers_to_calculate = [
+                    value["identifier"]
+                    for value in calculated_summary_block["calculation"]["operation"][
+                        "+"
+                    ]
+                    if value["source"] == "answers"
+                ]
```
MebinAbraham, line 426: This is essentially a duplicate of `get_calculated_answer_ids`. I suggest rewriting that to work with a given schema. Then you can reuse here.

schemas/blocks/calculated_summary.json
```diff
@@ -24,26 +24,55 @@
...
+                "properties": {
+                  "+": {
+                    "type": "array",
+                    "items": [
+                      {
+                        "$ref": "https://eq.ons.gov.uk/value_sources.json#/value_source_for_calculations",
+                        "description": "A value source to validate against that resolves to a number."
+                      }
+                    ]
+                  }
```
MebinAbraham, line 64: - This should be introduced as a new numeric operator as part of rules. We currently have boolean and value operators. This is just a new one.

- Rules should also be extended and tested
- The numeric operator should allow pretty much any source, for answers we need to check it is numeric. There is code validation to do such things already which should be extended.
- It needs to allow recursion of other numerical operators (currently only addition)

Think of it as you need to add a new `+` in general than calculated summary specifically. We just happen to reuse it for calculated summaries.

app/validators/routing/types.py
```diff
@@ -51,8 +51,12 @@ def resolve_value_source_json_type(value_source, schema):
...
 
     if source == "calculated_summary":
-        block_id = schema.get_block(value_source["identifier"])
-        answer_id = block_id["calculation"]["answers_to_calculate"][0]
+        block = schema.get_block(value_source["identifier"])
+        if block["calculation"].get("answers_to_calculate"):
+            answer_id = block["calculation"]["answers_to_calculate"][0]
+        else:
+            answer_value_source = block["calculation"]["operation"]["+"][0]
+            answer_id = answer_value_source["identifier"]
```
MebinAbraham, line 59: I can't remember why we only use the first answer?. Do we validate that answers for calculated summary must be the same type already?.

app/validators/blocks/calculated_summary_block_validator.py
berroar: Added the check

schemas/blocks/calculated_summary.json
berroar: Added the new numeric "+" operator 👍

app/validators/blocks/calculated_summary_block_validator.py
```diff
@@ -31,21 +37,44 @@ def validate(self):
...
-            ):
+            if any(answer["currency"] != answers[0]["currency"] for answer in answers):
                 self.add_error(self.ANSWERS_MUST_HAVE_SAME_CURRENCY)
         return self.errors
+
+    def validate_answer_id_set_before_calculated_summary_block(self):
+        for answer_id in self.answers_to_calculate:
+            answer_id_block = self.questionnaire_schema.get_block_by_answer_id(
+                answer_id
+            )
```
MebinAbraham, line 63: If you only care about the block_id then you can use `get_block_id_by_answer_id`.

schemas/blocks/calculated_summary.json
```diff
@@ -24,26 +24,47 @@
...
+          {
+            "type": "object",
+            "properties": {
+              "calculation_type": {
+                "type": "string",
+                "enum": ["sum"]
+              },
+              "answers_to_calculate": {
+                "type": "array",
+                "minItems": 2,
```
MebinAbraham, line 37: I suggest, "uniqueItems": true"` also.

tests/validators/rules/test_rules_validator.py
```diff
@@ -215,6 +215,81 @@ def test_validate_count_operator_non_checkbox_answer():
...
+        "origin_id": "block-id",
+    }
+
+    assert validator.errors == [expected_error]
+
+
+def test_validate_nested_sum_operator():
+    sum_operator = {
+        "+": [
+            {"+": [{"source": "answers", "identifier": "array-answer"}]},
```
MebinAbraham, line 243: Array length can't be one for the numeric operator. Need at least 2 things to add.

tests/validators/rules/test_rules_validator.py
```diff
@@ -215,6 +215,81 @@ def test_validate_count_operator_non_checkbox_answer():
...
+            {"+": [{"source": "answers", "identifier": "checkbox-answer"}, ["Yes"]]},
```
MebinAbraham, line 244: This doesn't make sense, what is the "Yes" for?.

MebinAbraham (commented):
Need to add new schemas to `tests/schemas/rules/`

app/validators/rules/rule_validator.py
```diff
@@ -237,6 +250,27 @@ def _validate_count_operator(self, operator):
                 value_source=first_argument,
             )
 
+    def _validate_sum_operator(self, operator):
```
MebinAbraham, line 253: Let's discuss offline, we actually don't need to validate these operators seperately, it is covered in `new_when_rule_validator.py`. It just needs extending and porting since operator can be used outside of when rules now.

berroar: I have updated this now as discussed 👍

app/validators/rules/rule_validator.py
```diff
@@ -262,3 +271,97 @@ def _validate_options(self, rules, operator_name):
...
+
+        if operator_name in [Operator.ANY_IN, Operator.ALL_IN]:
+            return [TYPE_ARRAY]
+
+        if operator_name == Operator.IN:
+            return (
+                [TYPE_NUMBER, TYPE_STRING] if argument_position == 0 else [TYPE_ARRAY]
+            )
+        if operator_name in [Operator.COUNT, Operator.SUM]:
+            return [TYPE_NUMBER]
```
MebinAbraham, line 367: This should return the valid types for an operators arguments. For count it is an array.
```py
        if operator_name == Operator.COUNT:
            return [TYPE_ARRAY]

        if operator_name == Operator.SUM:
            return [TYPE_NUMBER]
```

app/validators/rules/rule_validator.py
```diff
@@ -262,3 +271,97 @@ def _validate_options(self, rules, operator_name):
...
+        ):
+            self._validate_argument_types_match(rule, argument_types)
+
+        if operator_name == Operator.DATE:
+            return TYPE_DATE
+
+        if operator_name == Operator.COUNT:
+            return TYPE_NUMBER
+
+        return TYPE_BOOLEAN
```
MebinAbraham, line 296: Need one for `NUMERIC` operators instead of COUNT.

berroar: Updated 👍

app/validators/rules/rule_validator.py
berroar: Updated 👍

app/validators/questionnaire_schema.py
```diff
@@ -435,3 +452,20 @@ def _get_numeric_value(self, defined_value, system_default, answer_ranges):
...
+
+    @staticmethod
+    def get_calculated_answer_ids(block):
+        if block["calculation"].get("answers_to_calculate"):
+            return block["calculation"]["answers_to_calculate"]
+
+        values = get_values_for_key(block["calculation"]["operation"], "+")
+
+        calculated_summary_answer_ids = []
+        for value_sources in values:
```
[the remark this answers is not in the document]
berroar: Updated to use this method, and updated the valid calculated summary schema to use a nested operator 👍

tests/validators/rules/test_rules_validator.py
berroar: This has been updated now 👍

app/validators/rules/rule_validator.py
```diff
@@ -237,6 +250,27 @@ def _validate_count_operator(self, operator):
                 value_source=first_argument,
             )
 
+    def _validate_sum_operator(self, operator):
+        """
+        Validates that an answer value source within a sum operator is numeric
+        """
+        values = get_values_for_key(operator, "+")
```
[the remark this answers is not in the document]
berroar: This has been removed now 👍

app/validators/blocks/calculated_summary_block_validator.py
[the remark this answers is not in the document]
berroar: Updated this

app/validators/blocks/calculated_summary_block_validator.py
```diff
@@ -31,21 +37,44 @@ def validate(self):
...
+
+    def validate_answer_id_for_calculated_summary_not_in_different_section(self):
+        answer_section_ids = {
+            self.questionnaire_schema.answers_with_context[answer_id]["section"]
+            for answer_id in self.answers_to_calculate
+        }
+        if len(answer_section_ids) > 1:
+            self.add_error(
+                self.ANSWER_SET_IN_DIFFERENT_SECTION_FOR_CALCULATED_SUMMARY,
+                block=self.block,
```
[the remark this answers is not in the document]
berroar: Updated. 👍

schemas/blocks/calculated_summary.json
berroar: Added 👍

schemas/blocks/calculated_summary.json
```diff
@@ -24,26 +24,48 @@
...
-            "$ref": "https://eq.ons.gov.uk/string_interpolation/definitions.json#/string_with_placeholders"
+          {
+            "type": "object",
+            "properties": {
+              "operation": {
+                "type": "object",
+                "properties": {
+                  "$ref": "https://eq.ons.gov.uk/rules/operators/numeric/sum.json#/sum_answer_values"
+                },
+                "required": ["+"]
```
MebinAbraham, line 59: This will need `"additionalProperties": false`

schemas/rules/operators/numeric/sum.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://eq.ons.gov.uk/rules/operators/value/sum.json",
+  "description": "Returns the sum of items in an array. Accepts a single argument which is the array",
```
MebinAbraham, line 4: This is now incorrect.

tests/validators/rules/test_rules_validator.py
```diff
@@ -207,14 +207,84 @@ def test_validate_count_operator_non_checkbox_answer():
...
+        "operator": "count",
+        "origin_id": "block-id",
+        "valid_types": ["array"],
     }
 
     assert validator.errors == [expected_error]
 
 
+def test_validate_sum_operator():
+    sum_operator = {"+": [{"source": "answers", "identifier": "array-answer"}]}
```
MebinAbraham, line 222: Although this is a test, we should use realistic valid format, so I would add another answer source or number to this, so the array is 2 in length.

schemas/blocks/calculated_summary.json
berroar: As discussed needs to be resolved in 2020 json schema draft

schemas/rules/operators/numeric/sum.json
berroar: Updated 👍

app/validators/routing/types.py
berroar: Yeah I believe so, we made this decision when we added validation for calculated summary source types in when rules.

MebinAbraham (approved):

rmccar (approved):

PR #168, opened 2023-01-09
Add preview questions rules

petechd:
### PR Context
Adds new schema field (`preview_questions`) and extra python validation. `if/else` validation in schema definitions proved to be problematic as `questionnaire_flow` where you extract `type`: `Linear`/`Hub` is nested on different level than `sections`. This PR is required for the preview question runner changes [PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/996).

### Checklist


petechd:
[Viewing questions before starting the survey - Iteration 1 (5 SP)](https://trello.com/c/BipDuXxj/5391-viewing-questions-before-starting-the-survey-iteration-1-5-sp)

rmccar (approved):

Yuyuutsu (approved):
LGTM !

app/validators/questionnaire_validator.py
```diff
@@ -94,3 +99,19 @@ def validate_smart_quotes(self):
...
+
+    def validate_introduction_block(self):
+        blocks = self.questionnaire_schema.get_blocks()
+        has_introduction_blocks = any(
+            block["type"] == "Introduction" for block in blocks
+        )
+        if not has_introduction_blocks:
+            self.add_error(error_messages.NO_INTRODUCTION_BLOCK)
+
+    def validate_section_title(self):
```
MebinAbraham, line 111: This should be done via JSON schema validation.

Hajara-I:
Looks good to me

app/validators/questionnaire_validator.py
petechd: I tried but since we need `questionnaire_flow` field and section's nested attributes is it possible in JSON schema validation? I can't find similar example.

schemas/questionnaire_v1.json
```diff
@@ -402,6 +402,10 @@
         },
         "required": ["id", "groups"]
       }
+    },
+    "preview_questions": {
+      "type": "boolean",
+      "description": "A flag that is used to determine if the questions are previewed before the questionnaire start"
```
MebinAbraham, line 408: Why did we need an introduction page?. If it is specific to introduction page, then we should make this flag a property of Introduction block.

```suggestion
      "description": "A flag that is used to determine if the questions can be previewed before the questionnaire start"
```

app/error_messages.py
```diff
@@ -37,3 +37,5 @@
 VARIANTS_HAVE_MULTIPLE_QUESTION_TYPES = (
     "Variants have more than one question type for block."
 )
+NO_INTRODUCTION_BLOCK = "No introduction block for preview questions in schema"
+NO_SECTION_TITLE = "No section title for section in schema using hub"
```
MebinAbraham, line 41: Can be removed.

app/error_messages.py
```diff
@@ -37,3 +37,5 @@
 VARIANTS_HAVE_MULTIPLE_QUESTION_TYPES = (
     "Variants have more than one question type for block."
 )
+NO_INTRODUCTION_BLOCK = "No introduction block for preview questions in schema"
```
MebinAbraham, line 40: This should be more meaningful, maybe something like `No introduction block found. Introduction block is mandatory when using the preview questions feature.`

app/validators/questionnaire_validator.py
```diff
@@ -104,3 +107,11 @@ def validate_smart_quotes(self):
                         error_messages.DUMB_QUOTES_FOUND,
                         pointer=translatable_item.pointer,
                     )
+
+    def validate_introduction_block(self):
```
MebinAbraham, line 111: Needs a test

petechd: Added a test, thanks!

app/error_messages.py
petechd: Removed.

app/error_messages.py
petechd: Added.

app/error_messages.py
```diff
@@ -37,3 +37,4 @@
 VARIANTS_HAVE_MULTIPLE_QUESTION_TYPES = (
     "Variants have more than one question type for block."
 )
+NO_INTRODUCTION_BLOCK = "No introduction block found. Introduction block is mandatory when using the preview questions feature."
```
MebinAbraham, line 40: Minor, can be better named.
```suggestion
PREVIEW_WITHOUT_INTRODUCTION_BLOCK = "No introduction block found. Introduction block is mandatory when using the preview questions feature."
```

schemas/questionnaire_v1.json
```diff
@@ -398,8 +398,25 @@
...
+              "properties": { "type": { "const": "Linear" } }
+            }
+          }
+        },
+        "then": {
+          "required": ["id", "groups"]
+        },
+        "else": {
+          "required": ["id", "groups", "title"]
+        }
```
MebinAbraham, line 414: This doesn't seem to work. A Hub survey without a title is not passing for me.

petechd: It's because `if` conditions keyword is not found in schema so `then` is being used, it might be impossible to do this logic in JSON schema definitions.

MebinAbraham: It is possible. It is currently defined incorrect at the wrong level, JSON schema can't look upwards to a parent's value. So this need to set `"required": ["id", "groups"]` as this is always true.

Then at the top level you need:
```json
"allOf": [
    {
      "if": {
        "properties": {
          "questionnaire_flow": {
            "properties": {
              "type": {
                "const": "Hub"
              }
            }
          }
        }
      },
      "then": {
        "properties": {
          "sections": {
            "items": {
              "required": ["title"]
            }
          }
        }
      }
    }
  ]
```

petechd: Thanks, I used it and it works like a charm!

rmccar (approved):

PR #169, opened 2023-01-12
Add BEIS-NI theme

rmccar:
### PR Context
Adds the ability to set BEIS-NI theme in runner

### Checklist

### DIFF
--- a/app/validators/metadata_validator.py
+++ b/app/validators/metadata_validator.py
@@ -32,7 +32,7 @@ class MetadataValidator(Validator):
         # ru_name required for template rendering in business, default and NI theme
         required_metadata_names = []
 
-        if self.theme_name in ["business", "default", "northernireland", "beis", "orr"]:
+        if self.theme_name in ["business", "default", "northernireland", "beis", "nibeis", "orr"]:
             required_metadata_names.extend(["user_id", "period_id", "ru_name"])
 
         for metadata_name in required_metadata_names:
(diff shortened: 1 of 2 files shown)


rmccar:
[Create runner theme for NI BEIS](https://trello.com/c/ODxARf77/5533-create-runner-theme-for-ni-beis)

app/validators/metadata_validator.py
```diff
@@ -32,7 +32,14 @@ def validate_mandatory(self):
         # ru_name required for template rendering in business, default and NI theme
         required_metadata_names = []
 
-        if self.theme_name in ["business", "default", "northernireland", "beis", "orr"]:
+        if self.theme_name in [
+            "business",
+            "default",
+            "northernireland",
+            "beis",
+            "nibeis",
```
MebinAbraham, line 40: To be consistent with census one in runner, this should be like `beis-ni`. Use a hyphen to make it clear what it is. `beis-northern-ireland` would be better for readability.

rmccar: In that case would you also want `northernireland` renamed to `northern-ireland` or `NI` as well?

MebinAbraham: No, only because they are live surveys using "northernireland", so you can't rename it without those changing. Ideally yes but we cant right now.

MebinAbraham (approved):

Yuyuutsu (approved):

PR #173, opened 2023-01-31
Add Support For Answer Codes

berroar:
### PR Context

Adds support for Answer Codes, see the design proposal in this [document](https://github.com/ONSdigital/eq-questionnaire-validator/blob/master/doc/decisions/0011-user-defined-answer-codes.md)

Also removes validation enforcing that answer labels and values match in answer options.

### Checklist

Check that the validation requirements outlined in the design proposal and the trello card has been covered by this PR.

Is any validation missing.

### DIFF
--- a/app/validators/answer_code_validator.py
+++ b/app/validators/answer_code_validator.py
@@ -176,4 +176,5 @@ class AnswerCodeValidator(Validator):
             self.add_error(
                 self.INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS,
                 answer_codes_for_options=answer_codes_for_options,
+                allowed_values=values,
             )
--- a/app/validators/answer_code_validator.py
+++ b/app/validators/answer_code_validator.py
@@ -79,7 +79,8 @@ class AnswerCodeValidator(Validator):
                 ]
 
                 if len(values) != len(answer_codes_for_options) and all(
-                    "value" in answer_code for answer_code in answer_codes_for_options
+                    "answer_value" in answer_code
+                    for answer_code in answer_codes_for_options
                 ):
                     self.add_error(
                         self.ANSWER_CODE_MISSING_FOR_ANSWER_OPTIONS,
@@ -88,15 +89,18 @@ class AnswerCodeValidator(Validator):
                     )
 
                 for answer_code in answer_codes_for_options:
-                    if answer_code.get("value") and answer_code["value"] not in values:
+                    if (
+                        answer_code.get("answer_value")
+                        and answer_code["answer_value"] not in values
+                    ):
                         self.add_error(
                             self.INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS,
-                            value=answer_code["value"],
+                            answer_value=answer_code["answer_value"],
                             answer_codes_for_options=answer_codes_for_options,
                         )
 
                 if len(answer_codes_for_options) != 1 and any(
-                    "value" not in answer_code
+                    "answer_value" not in answer_code
                     for answer_code in answer_codes_for_options
                 ):
                     # Multiple answer codes for options are allowed where options are dynamic
--- a/app/validators/answer_code_validator.py
+++ b/app/validators/answer_code_validator.py
@@ -16,6 +16,9 @@ class AnswerCodeValidator(Validator):
     INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS = (
         "Values specified in answer code and answer options do not match"
     )
+    DYNAMIC_ANSWER_OPTION_MUST_HAVE_ANSWER_CODE_SET_AT_TOP_LEVEL = (
+        "Dynamic options must have an answer code set at the parent level"
+    )
 
     def __init__(self, data_version, answer_codes, questionnaire_schema):
         self.data_version = data_version
@@ -34,6 +37,7 @@ class AnswerCodeValidator(Validator):
         self.validate_missing_answer_id()
         self.validate_missing_answer_codes()
         self.validate_answer_codes_at_option_level()
+        self.validate_dynamic_options()
         return self.errors
 
     def validate_data_version(self):
@@ -59,11 +63,30 @@ class AnswerCodeValidator(Validator):
             if answer_id not in self.answer_codes_answer_ids:
                 self.add_error(self.MISSING_ANSWER_CODE, answer_id=answer_id)
 
-    def validate_answer_codes_at_option_level(self):
-        answers_with_context = self.questionnaire_schema.answers_with_context
+    def validate_dynamic_options(self):
+        for answer_id in self.questionnaire_schema.answers_with_context:
+            answer = self.questionnaire_schema.answers_with_context[answer_id]
 
-        for answer_id in answers_with_context:
-            answer = answers_with_context[answer_id]
+            if "dynamic_options" in answer["answer"]:
+                answer_codes_for_options = [
+                    answer_code
+                    for answer_code in self.answer_codes
+                    if answer_code["answer_id"] == answer_id
+                ]
+
+                top_level_answer_code_count = sum(
+                    "answer_value" not in answer_code
+                    for answer_code in answer_codes_for_options
+                )
+                if top_level_answer_code_count == 0:
+                    self.add_error(
+                        self.DYNAMIC_ANSWER_OPTION_MUST_HAVE_ANSWER_CODE_SET_AT_TOP_LEVEL,
+                        answer_codes_for_options=answer_codes_for_options,
+                    )
+
+    def validate_answer_codes_at_option_level(self):
+        for answer_id in self.questionnaire_schema.answers_with_context:
+            answer = self.questionnaire_schema.answers_with_context[answer_id]
 
             if "options" in answer["answer"]:
                 values = []
@@ -111,9 +134,9 @@ class AnswerCodeValidator(Validator):
                                 answer_codes_for_options=answer_codes_for_options,
                             )
 
-                    elif not answers_with_context[answer_id]["answer"].get(
-                        "dynamic_options"
-                    ):
+                    elif not self.questionnaire_schema.answers_with_context[answer_id][
+                        "answer"
+                    ].get("dynamic_options"):
                         self.add_error(
                             self.MORE_THAN_ONE_ANSWER_CODE_SET_FOR_ANSWER_OPTIONS,
                             answer_options=answer["answer"]["options"],
--- a/app/validators/answer_code_validator.py
+++ b/app/validators/answer_code_validator.py
@@ -1,0 +1,106 @@
+from app.validators.questionnaire_schema import find_duplicates
+from app.validators.validator import Validator
+
+
+class AnswerCodeValidator(Validator):
+    INCORRECT_DATA_VERSION_FOR_ANSWER_CODES = (
+        "Answer codes are only supported in data version 0.0.3"
+    )
+    DUPLICATE_ANSWER_CODE_FOUND = "Answer codes must be unique"
+    MISSING_ANSWER_CODE = "No answer codes found for answer_id set in the schema"
+    ANSWER_CODE_ANSWER_ID_NOT_FOUND_IN_SCHEMA = (
+        "No matching answer id found in the schema for the given answer code"
+    )
+    ANSWER_CODE_MISSING_FOR_ANSWER_OPTIONS = "Number of answer codes does not match number of answers for an answer with multiple options"
+    MORE_THAN_ONE_ANSWER_CODE_SET_FOR_ANSWER_OPTIONS = "Only one answer code should be set for answers with answer options if not specifying a value"
+    INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS = (
+        "Values specified in answer code and answer options do not match"
+    )
+
+    def __init__(self, data_version, answer_codes, questionnaire_schema):
+        self.data_version = data_version
+        self.answer_codes = answer_codes
+        self.questionnaire_schema = questionnaire_schema
+        self.codes = [answer["code"] for answer in self.answer_codes]
+        self.answer_codes_answer_ids = {
+            answer["answer_id"] for answer in self.answer_codes
+        }
+        super().__init__(questionnaire_schema)
+
+    def validate(self):
+        self.validate_data_version()
+        self.validate_duplicate_answer_codes()
+        self.validate_missing_answer_id()
+        self.validate_missing_answer_codes()
+        self.validate_answer_codes_at_option_level()
+        return self.errors
+
+    def validate_data_version(self):
+        if self.data_version != "0.0.3":
+            self.add_error(self.INCORRECT_DATA_VERSION_FOR_ANSWER_CODES)
+
+    def validate_duplicate_answer_codes(self):
+        duplicates = find_duplicates(self.codes)
+
+        if len(duplicates) > 0:
+            self.add_error(self.DUPLICATE_ANSWER_CODE_FOUND, duplicates=duplicates)
+
+    def validate_missing_answer_id(self):
+        all_answer_ids = list(self.questionnaire_schema.answers_with_context)
+
+        for answer_code_id in self.answer_codes_answer_ids:
+            if answer_code_id not in all_answer_ids:
+                self.add_error(
+                    self.ANSWER_CODE_ANSWER_ID_NOT_FOUND_IN_SCHEMA,
+                    answer_id=answer_code_id,
+                )
+
+    def validate_missing_answer_codes(self):
+        all_answer_ids = list(self.questionnaire_schema.answers_with_context)
+
+        for answer_id in all_answer_ids:
+            if answer_id not in self.answer_codes_answer_ids:
+                self.add_error(self.MISSING_ANSWER_CODE, answer_id=answer_id)
+
+    def validate_answer_codes_at_option_level(self):
+        answers_with_context = self.questionnaire_schema.answers_with_context
+
+        for answer_id in answers_with_context:
+            answer = answers_with_context[answer_id]
+
+            if "options" in answer["answer"]:
+                values = []
+                values.extend(option["value"] for option in answer["answer"]["options"])
+
+                answer_codes_for_options = [
+                    answer_code
+                    for answer_code in self.answer_codes
+                    if answer_code["answer_id"] == answer_id
+                ]
+
+                if len(values) != len(answer_codes_for_options) and all(
+                    "value" in answer_code for answer_code in answer_codes_for_options
+                ):
+                    self.add_error(
+                        self.ANSWER_CODE_MISSING_FOR_ANSWER_OPTIONS,
+                        answer_options=answer["answer"]["options"],
+                        answer_codes_for_options=answer_codes_for_options,
+                    )
+
+                for answer_code in answer_codes_for_options:
+                    if answer_code.get("value") and answer_code["value"] not in values:
+                        self.add_error(
+                            self.INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS,
+                            value=answer_code["value"],
+                            answer_codes_for_options=answer_codes_for_options,
+                        )
+
+                if len(answer_codes_for_options) != 1 and any(
+                    "value" not in answer_code
+                    for answer_code in answer_codes_for_options
+                ):
+                    self.add_error(
+                        self.MORE_THAN_ONE_ANSWER_CODE_SET_FOR_ANSWER_OPTIONS,
+                        answer_options=answer["answer"]["options"],
+                        answer_codes_for_options=answer_codes_for_options,
+                    )
--- a/app/validators/answer_code_validator.py
+++ b/app/validators/answer_code_validator.py
@@ -1,3 +1,4 @@
+from app.validators.questionnaire_schema import find_duplicates
 from app.validators.validator import Validator
 
 
@@ -6,6 +7,7 @@ class AnswerCodeValidator(Validator):
         "Answer codes are only supported in data version 0.0.3"
     )
     DUPLICATE_ANSWER_CODE_FOUND = "Answer codes must be unique"
+    DUPLICATE_ANSWER_ID_FOUND = "Answer ids must only have one answer code unless answer codes are being set against answer values"
     MISSING_ANSWER_CODE = "No answer codes found for answer_id set in the schema"
     ANSWER_CODE_ANSWER_ID_NOT_FOUND_IN_SCHEMA = (
         "No matching answer id found in the schema for the given answer code"
@@ -48,19 +50,25 @@ class AnswerCodeValidator(Validator):
             self.add_error(self.INCORRECT_DATA_VERSION_FOR_ANSWER_CODES)
 
     def validate_duplicate_answer_codes(self):
-        duplicate_answer_ids = [
-            answer_code["answer_id"]
-            for answer_code in self.answer_codes
-            if "answer_value" not in answer_code
-        ]
-        if duplicates := {
-            answer_id
-            for answer_id in duplicate_answer_ids
-            if duplicate_answer_ids.count(answer_id) > 1
-        }:
-            self.add_error(
-                self.DUPLICATE_ANSWER_CODE_FOUND, duplicate_answer_ids=duplicates
-            )
+        duplicates = find_duplicates(self.codes)
+
+        if len(duplicates) > 0:
+            self.add_error(self.DUPLICATE_ANSWER_CODE_FOUND, duplicates=duplicates)
+        else:
+            duplicate_answer_ids = [
+                answer_code["answer_id"]
+                for answer_code in self.answer_codes
+                if "answer_value" not in answer_code
+            ]
+            if duplicate_ids := {
+                answer_id
+                for answer_id in duplicate_answer_ids
+                if duplicate_answer_ids.count(answer_id) > 1
+            }:
+                self.add_error(
+                    self.DUPLICATE_ANSWER_ID_FOUND,
+                    duplicate_answer_ids=list(duplicate_ids),
+                )
 
     def validate_missing_answer_id(self):
         for answer_code_id in self.answer_codes_answer_ids:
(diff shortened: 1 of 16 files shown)


berroar:
[Implement answer_codes in runner and validator (5 SP)](https://trello.com/c/5k8NixXD/5515-implement-answercodes-in-runner-and-validator-5-sp)

schemas/questionnaire_v1.json
```diff
@@ -71,6 +71,31 @@
...
+            "pattern": "^[a-zA-Z0-9-_]+$",
+            "description": "User defined answer code for a matching answer id"
+          },
+          "answer_id": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+            "description": "Answer id of the answer code"
+          },
+          "value": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+            "description": "Answer id of the answer code"
```
MebinAbraham, line 92: This should be `answer_value` as per the docs. Also, the description needs updating.

schemas/questionnaire_v1.json
```diff
@@ -71,6 +71,31 @@
...
+    "answer_codes": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "items": {
+        "type": "object",
+        "properties": {
+          "code": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+            "pattern": "^[a-zA-Z0-9-_]+$",
```
MebinAbraham, line 83: This should match the regex in the JSON schema definition on ons-schema-definitions. I think `dots` seperations missing.

schemas/questionnaire_v1.json
```diff
@@ -71,6 +71,31 @@
...
+            "description": "User defined answer code for a matching answer id"
+          },
+          "answer_id": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
```
MebinAbraham, line 87:
```suggestion
            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/identifier",
```

MebinAbraham (commented):
Haven't looked at validation yet but worth sorting these in the meantime.

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,110 @@
...
+                        self.add_error(
+                            self.INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS,
+                            value=answer_code["value"],
+                            answer_codes_for_options=answer_codes_for_options,
+                        )
+
+                if len(answer_codes_for_options) != 1 and any(
+                    "value" not in answer_code
+                    for answer_code in answer_codes_for_options
+                ):
```
MebinAbraham, line 101: This doesn't cover the case where you have 1 code for options but it also defines "answer_value".

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,110 @@
...
+                for answer_code in answer_codes_for_options:
+                    if answer_code.get("value") and answer_code["value"] not in values:
+                        self.add_error(
+                            self.INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS,
+                            value=answer_code["value"],
+                            answer_codes_for_options=answer_codes_for_options,
+                        )
+
+                if len(answer_codes_for_options) != 1 and any(
+                    "value" not in answer_code
```
MebinAbraham, line 99: This needs to change to `answer_value`

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,110 @@
...
+                    for answer_code in answer_codes_for_options
+                ):
+                    # Multiple answer codes for options are allowed where options are dynamic
+                    if not answers_with_context[answer_id]["answer"].get(
+                        "dynamic_options"
+                    ):
+                        self.add_error(
+                            self.MORE_THAN_ONE_ANSWER_CODE_SET_FOR_ANSWER_OPTIONS,
```
MebinAbraham, line 107: I found this hard to resolve, it is not clear on what is needed to fix the issue. Based on the current logic this is more often true when you forget to define answer_value for some option. It might be better to say, answer code with matching answer_value not found for option. Worth also adding something like "Either all option values must have a code or none of them should".

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,110 @@
...
+                )
+
+    def validate_missing_answer_codes(self):
+        all_answer_ids = list(self.questionnaire_schema.answers_with_context)
+
+        for answer_id in all_answer_ids:
+            if answer_id not in self.answer_codes_answer_ids:
+                self.add_error(self.MISSING_ANSWER_CODE, answer_id=answer_id)
+
+    def validate_answer_codes_at_option_level(self):
```
MebinAbraham, line 65: Needs to add validation to ensure, if dynamic options are present then a code without `answer_value` is defined. At the moment, if you have dynamic and static options, if all static options have a code mapping, this still passes validation even though the dynamic option does have a code.

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,110 @@
...
+            self.add_error(self.INCORRECT_DATA_VERSION_FOR_ANSWER_CODES)
+
+    def validate_duplicate_answer_codes(self):
+        duplicates = find_duplicates(self.codes)
+
+        if len(duplicates) > 0:
+            self.add_error(self.DUPLICATE_ANSWER_CODE_FOUND, duplicates=duplicates)
+
+    def validate_missing_answer_id(self):
+        all_answer_ids = list(self.questionnaire_schema.answers_with_context)
```
MebinAbraham, line 49: I would define this at the object level since it is used multiple times.

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def validate_answer_codes_at_option_level(self):
+        answers_with_context = self.questionnaire_schema.answers_with_context
+
+        for answer_id in answers_with_context:
+            answer = answers_with_context[answer_id]
+
+            if "options" in answer["answer"]:
+                values = []
+                values.extend(option["value"] for option in answer["answer"]["options"])
```
MebinAbraham, line 73: At the moment, any arbitrary values can be defined for answer codes.
- The values from an answer code must exist in the schema
- The values from options must exist in answer code (if options should define a code)

schemas/questionnaire_v1.json
berroar: Updated these

schemas/questionnaire_v1.json
berroar: Updated now 👍

schemas/questionnaire_v1.json
berroar: added 👍

app/validators/answer_code_validator.py
berroar: Updated these 👍

app/validators/answer_code_validator.py
berroar: Updated the logic here and added another test 👍

app/validators/answer_code_validator.py
berroar: Updated this 👍

app/validators/answer_code_validator.py
berroar: Validating now that there is a top level code defined 👍

app/validators/answer_code_validator.py
berroar: Slightly confused on this one, do you mean not just in options? I thought the error for `INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS`  was covering this

app/validators/answer_code_validator.py
berroar: I think the name of this one was causing confusion - I added this to test that multiple answer codes could only be set if answer options were dynamic. Updated the name of this to reflect this 👍 I have also added a new check and test to cover the scenario that is more like the description of this error, where there are two answer codes with the same value but different codes.

schemas/questionnaire_v1.json
```diff
@@ -71,6 +71,31 @@
...
+            "pattern": "^[a-zA-Z0-9._-]+$",
+            "description": "User defined answer code for a matching answer id"
+          },
+          "answer_id": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/identifier",
+            "description": "Answer id of the answer code"
+          },
+          "answer_value": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+            "description": "Optional answer value for the given answer code"
```
MebinAbraham, line 92:
```suggestion
            "description": "Optional answer value for the given answer code. This should only used for answers that support `options`."
```

schemas/questionnaire_v1.json
```diff
@@ -71,6 +71,31 @@
...
+        "type": "object",
+        "properties": {
+          "code": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/non_empty_string",
+            "pattern": "^[a-zA-Z0-9._-]+$",
+            "description": "User defined answer code for a matching answer id"
+          },
+          "answer_id": {
+            "$ref": "https://eq.ons.gov.uk/common_definitions.json#/identifier",
+            "description": "Answer id of the answer code"
```
MebinAbraham, line 88:
```suggestion
            "description": "The `answer_id` this `code` mapping is for."
```

app/validators/answer_code_validator.py
berroar: Added validation and a test now for the scenario where an answer value is set for an answer code but there are no options.

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,179 @@
...
+                        answer_codes_for_options=answer_codes_for_options,
+                    )
+
+        if len(answer_values) != len(values) and any(
+            "answer_value" in answer_code for answer_code in answer_codes_for_options
+        ):
+            self.add_error(
+                self.INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS,
+                answer_codes_for_options=answer_codes_for_options,
+            )
```
MebinAbraham, line 179: This is hard to diagnose for author. Error says value doesn't match but it is because they are different lengths. I suggest adding a new message or updating existing to `Values specified in answer code and answer options do not match or they are of different lengths.`

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,179 @@
...
+                        answer_value=answer_code["answer_value"],
+                        answer_codes_for_options=answer_codes_for_options,
+                    )
+
+        if len(answer_values) != len(values) and any(
+            "answer_value" in answer_code for answer_code in answer_codes_for_options
+        ):
+            self.add_error(
+                self.INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS,
+                answer_codes_for_options=answer_codes_for_options,
```
MebinAbraham, line 178:
```suggestion
                answer_codes_for_options=answer_codes_for_options,
                allowed_values=values,
```

MebinAbraham (commented):
Each option value must only have one answer code mapping. Currently, this would pass:

```json
        {
            "answer_id": "mandatory-checkbox-answer",
            "answer_value": "None",
            "code": "1aa"
        },
        {
            "answer_id": "mandatory-checkbox-answer",
            "answer_value": "None",
            "code": "1a"
        },
```

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,179 @@
...
+        answer_values = set()
+        for answer_code in answer_codes_for_options:
+            if answer_code.get("answer_value"):
+                answer_values.add(answer_code["answer_value"])
+
+                if answer_code["answer_value"] not in values:
+                    self.add_error(
+                        self.INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS,
+                        answer_value=answer_code["answer_value"],
+                        answer_codes_for_options=answer_codes_for_options,
```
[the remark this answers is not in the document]
berroar: Updated this 👍

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,180 @@
...
+        answer_values = set()
+        for answer_code in answer_codes_for_options:
+            if answer_code.get("answer_value"):
+                answer_values.add(answer_code["answer_value"])
+
+                if answer_code["answer_value"] not in values:
+                    self.add_error(
+                        self.INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS,
+                        answer_code=answer_code,
+                        allowed_values=answer_codes_for_options,
```
MebinAbraham, line 170: Allow values should be the actual options, i.e `values` var.

app/validators/answer_code_validator.py
berroar: Updated this 👍

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,179 @@
...
+
+        if len(duplicates) > 0:
+            self.add_error(self.DUPLICATE_ANSWER_CODE_FOUND, duplicates=duplicates)
+
+    def validate_missing_answer_id(self):
+        for answer_code_id in self.answer_codes_answer_ids:
+            if answer_code_id not in self.all_answer_ids:
+                self.add_error(
+                    self.ANSWER_CODE_ANSWER_ID_NOT_FOUND_IN_SCHEMA,
+                    answer_id=answer_code_id,
```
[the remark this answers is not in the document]
berroar: Added 👍

app/validators/answer_code_validator.py
berroar: Updated now 👍

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,180 @@
...
+                    answer_options=answer["answer"]["options"],
+                    answer_codes_for_options=answer_codes_for_options,
+                )
+
+    def validate_incorrect_values_in_answer_options(
+        self, answer_codes_for_options, values
+    ):
+        answer_values = set()
+        for answer_code in answer_codes_for_options:
+            if answer_code.get("answer_value"):
```
ONS-Guilhem-Forey, line 163:
```python
if answer_value := answer_code.get("answer_value"):
```

And then use `answer_value` below

app/validators/answers/option_answer_validator.py
```diff
@@ -90,21 +88,6 @@ def validate_duplicate_options(self):
...
-            if isinstance(option["label"], str):
-                label = option["label"]
-            else:
-                label = option["label"]["text"]
-
-            if label != option["value"]:
-                self.add_error(
-                    self.ANSWER_LABEL_VALUE_MISMATCH, label=label, value=option["value"]
-                )
-
```
ONS-Guilhem-Forey, line 107: What's the reason for removing this code?

ONS-Guilhem-Forey (commented):
Had a good look at the changes and the proposal.

Just one NIT comment and a question.

Other than that LGTM

app/validators/answer_code_validator.py
berroar: Added this 👍

app/validators/answers/option_answer_validator.py
berroar: This is one of the requirements on the card, to remove validation that enforces that answer values match answer labels. Basically it is to cover this part of the design proposal:
`For answers with options, an optional answer_value must be provided to allow mapping to a single option value. The validation that enforces the label must match the value will be removed to prevent downstream mappings from changing when the label changes.`

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,197 @@
...
+                        answer_codes_for_options=answer_codes_for_options,
+                    )
+
+    def validate_answer_codes_at_option_level(self):
+        for answer_id in self.questionnaire_schema.answers_with_context:
+            answer = self.questionnaire_schema.answers_with_context[answer_id]
+
+            block = self.questionnaire_schema.get_block_by_answer_id(answer_id)
+
+            if block["type"] in ["ListEditQuestion", "ListRemoveQuestion"]:
```
MebinAbraham, line 108: Need to handle primary person block since we support that in runner. They are named slightly differently for primary.

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,197 @@
...
+                            answer_code=answer_code,
+                        )
+
+    def validate_missing_answer_codes_for_answer_options(
+        self, answer, answer_id, answer_codes_for_options, values
+    ):
+        if len(values) != len(answer_codes_for_options) and all(
+            "answer_value" in answer_code for answer_code in answer_codes_for_options
+        ):
+            self.add_error(
```
MebinAbraham, line 147: I would add `answer_id` to the error.

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,197 @@
...
+                answer_codes_for_options=answer_codes_for_options,
+            )
+
+        if any(
+            "answer_value" not in answer_code
+            for answer_code in answer_codes_for_options
+        ):
+            if len(answer_codes_for_options) == 1:
+                if len(values) != 1 and "answer_value" in answer_codes_for_options[0]:
+                    self.add_error(
```
MebinAbraham, line 159: I would add `answer_id` to the error.

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,197 @@
...
+    ANSWER_VALUE_SET_FOR_ANSWER_WITH_NO_OPTIONS = (
+        "Answer values can only be set for answers that support answer options"
+    )
+    ANSWER_CODE_MISSING_FOR_ANSWER_OPTIONS = (
+        "The number of answer codes does not match number of answer options"
+    )
+    MORE_THAN_ONE_ANSWER_CODE_SET_AT_PARENT_LEVEL = "Only one answer code should be set for an answer when not specifying answer codes for answer options"
+    INCORRECT_VALUE_FOR_ANSWER_CODE_WITH_ANSWER_OPTIONS = "Values specified in answer code and answer options do not match or they are of different lengths."
+    DYNAMIC_ANSWER_OPTION_MUST_HAVE_ANSWER_CODE_SET_AT_TOP_LEVEL = (
+        "Dynamic options must have an answer code set at the parent level"
```
MebinAbraham, line 23: This isn't very clear on how to resolve. Maybe something like `Answers with dynamic options must have an answer code mapping without answer value`?. Var name needs to change accordingly. Parent level is confusing since the code isnt being defined at the answer level.

MebinAbraham (commented):
At the moment, you can define a multiple codes for the same `answer_id`. This currently doesn't raise an error but should:
i.e

```json
        {
            "answer_id": "mandatory-checkbox-answer",
            "code": "12"
        },
        {
            "answer_id": "mandatory-checkbox-answer",
            "code": "1"
        },
```

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,197 @@
...
+    ANSWER_CODE_ANSWER_ID_NOT_FOUND_IN_SCHEMA = (
+        "No matching answer id found in the schema for the given answer code"
+    )
+    ANSWER_VALUE_SET_FOR_ANSWER_WITH_NO_OPTIONS = (
+        "Answer values can only be set for answers that support answer options"
+    )
+    ANSWER_CODE_MISSING_FOR_ANSWER_OPTIONS = (
+        "The number of answer codes does not match number of answer options"
+    )
+    MORE_THAN_ONE_ANSWER_CODE_SET_AT_PARENT_LEVEL = "Only one answer code should be set for an answer when not specifying answer codes for answer options"
```
MebinAbraham, line 20: Still not convinced this is enough o author to work out the issue. Worth adding to the end, `Either all options must defined a code or none of them should.`

app/validators/answer_code_validator.py
berroar: Updated this 👍

app/validators/answer_code_validator.py
berroar: Updated 👍

app/validators/answer_code_validator.py
berroar: Updated 👍

berroar:
> At the moment, you can define a multiple codes for the same `answer_id`. This currently doesn't raise an error but should: i.e.
> 
> ```json
>         {
>             "answer_id": "mandatory-checkbox-answer",
>             "code": "12"
>         },
>         {
>             "answer_id": "mandatory-checkbox-answer",
>             "code": "1"
>         },
> ```

Fixed the logic for this as previously was only finding duplicates at the top level with the same code 👍

app/validators/answer_code_validator.py
```diff
@@ -51,10 +48,19 @@ def validate_data_version(self):
...
+            if "answer_value" not in answer_code
+        ]
+        if duplicates := {
+            answer_id
+            for answer_id in duplicate_answer_ids
+            if duplicate_answer_ids.count(answer_id) > 1
+        }:
+            self.add_error(
+                self.DUPLICATE_ANSWER_CODE_FOUND, duplicate_answer_ids=duplicates
+            )
```
MebinAbraham, line 63: Duplicate code checking no longer works.

app/validators/answer_code_validator.py
```diff
@@ -51,10 +48,19 @@ def validate_data_version(self):
...
+            for answer_code in self.answer_codes
+            if "answer_value" not in answer_code
+        ]
+        if duplicates := {
+            answer_id
+            for answer_id in duplicate_answer_ids
+            if duplicate_answer_ids.count(answer_id) > 1
+        }:
+            self.add_error(
+                self.DUPLICATE_ANSWER_CODE_FOUND, duplicate_answer_ids=duplicates
```
MebinAbraham, line 62: This needs to be a `list` or the JSON response errors as `set` is not serialisable.

app/validators/answer_code_validator.py
berroar: Updated 👍

app/validators/answer_code_validator.py
berroar: Converting to list now 👍

app/validators/answer_code_validator.py
```diff
@@ -48,19 +50,25 @@ def validate_data_version(self):
...
+            ]
+            if duplicate_ids := {
+                answer_id
+                for answer_id in duplicate_answer_ids
+                if duplicate_answer_ids.count(answer_id) > 1
+            }:
+                self.add_error(
+                    self.DUPLICATE_ANSWER_ID_FOUND,
+                    duplicate_answer_ids=list(duplicate_ids),
+                )
```
MebinAbraham, line 71: Do you need the else?. It can do both as you may have duplicates codes and answer ids?.

Also, maybe the method can now just be `validate_duplicates`

app/validators/answer_code_validator.py
```diff
@@ -48,19 +50,25 @@ def validate_data_version(self):
...
+                if "answer_value" not in answer_code
+            ]
+            if duplicate_ids := {
+                answer_id
+                for answer_id in duplicate_answer_ids
+                if duplicate_answer_ids.count(answer_id) > 1
+            }:
+                self.add_error(
+                    self.DUPLICATE_ANSWER_ID_FOUND,
+                    duplicate_answer_ids=list(duplicate_ids),
```
MebinAbraham, line 70: To be consistent with the code, it would call this key `duplicates`.

berroar: Updated 👍

app/validators/answer_code_validator.py
```diff
@@ -48,19 +50,25 @@ def validate_data_version(self):
...
+            duplicate_answer_ids = [
+                answer_code["answer_id"]
+                for answer_code in self.answer_codes
+                if "answer_value" not in answer_code
+            ]
+            if duplicate_ids := {
+                answer_id
+                for answer_id in duplicate_answer_ids
+                if duplicate_answer_ids.count(answer_id) > 1
+            }:
```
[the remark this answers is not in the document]
berroar: Updated 👍

app/validators/answer_code_validator.py
berroar: Using the else to avoid producing two errors for the same issue when there is a duplicate answer code. Updated the method name though 👍

app/validators/answer_code_validator.py
```diff
@@ -0,0 +1,209 @@
...
+        if self.data_version != "0.0.3":
+            self.add_error(self.INCORRECT_DATA_VERSION_FOR_ANSWER_CODES)
+
+    def validate_duplicates(self):
+        duplicates = find_duplicates(self.codes)
+
+        if len(duplicates) > 0:
+            self.add_error(self.DUPLICATE_ANSWER_CODE_FOUND, duplicates=duplicates)
+        else:
+            duplicate_answer_ids = [
```
MebinAbraham, line 58: This is just `answer_ids`.

app/validators/answer_code_validator.py
MebinAbraham: Okay I see but it might not always be the same issue, as you may have 2 duplicates codes, and 2 different mappings with 2 duplicate answer ids. If one code/answer is duplicate the JSON validation should stop it as you have `uniqueItems` true.

app/validators/answer_code_validator.py
berroar: Renamed

app/validators/answer_code_validator.py
berroar: OK updated now

app/validators/answer_code_validator.py
```diff
@@ -54,17 +54,17 @@ def validate_duplicates(self):
 
         if len(duplicates) > 0:
```
MebinAbraham, line 55: To be consistent with the one below, this could just be `if duplicates`. I would update the var names to be obvious.

I suggest simplifying to something like this, since the check is the same for both.

```py
    def validate_duplicates(self):
        answer_ids = [
            answer_code["answer_id"]
            for answer_code in self.answer_codes
            if "answer_value" not in answer_code
        ]

        for values, error_message in [
            (self.codes, self.DUPLICATE_ANSWER_CODE_FOUND),
            (answer_ids, self.DUPLICATE_ANSWER_ID_FOUND),
        ]:

            duplicates = find_duplicates(values)

            if duplicates:
                self.add_error(error_message, duplicates=list(duplicates))
```

berroar: Updated 👍

ONS-Guilhem-Forey (approved):
Lgtm

PR #174, opened 2023-02-02
Proposal for dynamic answers based on a list

rmccar:
### PR Context
Proposes the structure to support dynamic list answer options.

### Checklist


rmccar:
[Iteration 3 - Design Proposal for dynamically creating answers based on a list (5 SP)](https://trello.com/c/VwUvOzHY/5488-iteration-3-design-proposal-for-dynamically-creating-answers-based-on-a-list-5-sp)

doc/decisions/0012-repeating-answer-options.md
```diff
@@ -0,0 +1,259 @@
...
+
+### `dynamic_answers`
+
+- A new property within the `question` block that supports answers based on a list of items, answers, or arbitrary functions 
+  that return a sequence of values.
+- `dynamic_answers` will contain two top level properties, `values` and `answers`.
+- `dynamic_answers` can be used alongside the existing static `answers` property, allowing for static and dynamic answers 
+  on the same page.
+- `dynamic_answers` must have at least one but can contain multiple `answers` to be iterated over for each literal list item.
+- `dynamic_answers` will be limited to a single block i.e. they will only be displayed on one page.
```
ONS-Guilhem-Forey, line 32: Just to clarify my understanding:
- A question can have only one `dynamic_answers` object
- A all dynamic answers of a `dynamic_answers` object must be displayed within the same block
- A schema can have multiple `dynamic_answers` objects, as long as there is maximum one of these per question.

?

doc/decisions/0012-repeating-answer-options.md
```diff
@@ -0,0 +1,259 @@
...
+  
+#### `dynamic_answers.values`
+
+- This field will contain the `source` for the list of literal values to be repeated over.
+- The `values` property will follow existing patterns and allow either a `Value Source` or a `Value Operator` as per the schema examples below.
+- There must be at least one item in the list being used to generate dynamic answers.
```
ONS-Guilhem-Forey, line 38: What happens if the list is empty?

Yuyuutsu (commented):
had a read through and it makes sense to me 👍

doc/decisions/0012-repeating-answer-options.md
```diff
@@ -0,0 +1,259 @@
...
+          },
+          "decimal_places": 2
+        }
+      ]
+    }
+  }
+}
+```
+- For dynamic answers driven by answer value sources, for each list item, a unique identifier like the answer value will need to be appended to the `id` in the `dynamic_answers.answer` block e.g. `"id": "percentage-of-shopping-{value}`.
+- We would need to enforce that any value used to be appended to the prefix would need to be unique, and would have to be converted to lowercase to be consistent with a valid answer `id`.
```
petechd, line 138: Just a detail really, but could be slightly improved with something like:
```suggestion
- We would need to enforce that any value used to be appended to the prefix would need to be unique, and would have to be formatted (eg. hyphens, lowercase) to be consistent with a valid answer `id`.
```

doc/decisions/0012-repeating-answer-options.md
rmccar: This is on the Author guys to prevent with routing rules. It will fall over if there is dynamic_answers and an  empty list

ONS-Guilhem-Forey (approved):

creativeappdevamit (commented):
Looks good to me!

Hajara-I:
Looks good to me!

doc/decisions/0012-repeating-answer-options.md
```diff
@@ -0,0 +1,269 @@
...
+## Properties and uses:
+
+### `dynamic_answers`
+
+- A new property within the `question` block that supports answers based on a list of items, answers, or arbitrary functions 
+  that return a sequence of values.
+- `dynamic_answers` will contain two top level properties, `values` and `answers`.
+- `dynamic_answers` can be used alongside the existing static `answers` property, allowing for static and dynamic answers 
+  on the same page.
+- `dynamic_answers` must have at least one but can contain multiple `answers` to be iterated over for each literal list item.
```
MebinAbraham, line 31: This isn't clear to me.

doc/decisions/0012-repeating-answer-options.md
```diff
@@ -0,0 +1,269 @@
...
+- `dynamic_answers` will be limited to a single block i.e. they will only be displayed on one page.
+  
+#### `dynamic_answers.values`
+
+- This field will contain the `source` for the list of literal values to be repeated over.
+- The `values` property will follow existing patterns and allow either a `Value Source` or a `Value Operator` as per the schema examples below.
+- There must be at least one item in the list being used to generate dynamic answers.
```
MebinAbraham, line 38: Not true as it can be empty if static answers are provided.

doc/decisions/0012-repeating-answer-options.md
```diff
@@ -0,0 +1,269 @@
...
+  
+#### `dynamic_answers.answers`
+- This property allows a list of answer blocks, as there can be more than one dynamic answer for each literal list item.
+- The `answers` property should be able to support any of the properties that would be supported in a normal/static `answer` block.
+- All the properties within the `dynamic_answers.answers` block that can contain text displayed to the user will need to support placeholder transforms. For example, answer labels which currently do not allow placeholders will need to be extended to support this. 
+- The structure should be consistent with a regular static answer block.
+- For `dynamic_answers.answers` blocks, the answer `id` will simply be a prefix, and a unique identifier will need to be added to the prefix at runtime. This is required in order for us to be able to render the id in other parts of the questionnaire.
```
MebinAbraham, line 45: Worth clarify it is only a prefix for rendering, but for storage, it is the ID. Worth adding that, we will identify each answer using `list_item_id` even if it isn't a list collector.

MebinAbraham (commented):
Nothing major.

doc/decisions/0012-repeating-answer-options.md
```diff
@@ -0,0 +1,274 @@
...
+            "value": 100
+          },
+          "decimal_places": 2
+        }
+      ]
+    }
+  }
+}
+```
+- For dynamic answers driven by list collectors, for front end purposes each list item, at runtime the `list-item-id` or `selector` will be appended to the `id` used in the `dynamic_answers.answer` block, so that all answers have unique ids e.g. `"id": "percentage-of-shopping-{list_item_item_id}`. For storage we will just use the `list_item_id`.
```
MebinAbraham, line 90:
```suggestion
- For dynamic answers driven by list collectors, for front end purposes each list item, at runtime the `list-item-id` will be appended to the `id` used in the `dynamic_answers.answer` block, so that all answers have unique ids e.g. `"id": "percentage-of-shopping-{list_item_id}`. For storage we will just use the combination of `answer_id` and `list_item_id`.
```

doc/decisions/0012-repeating-answer-options.md
```diff
@@ -0,0 +1,274 @@
...
+          "decimal_places": 2
+        }
+      ]
+    }
+  }
+}
+```
+
+- `list_item_id` can now be an answer value or the output of a function, not just list collector ids.
+- Dynamic answers can only be generated if there is at least one answer value.
```
MebinAbraham, line 141:
```suggestion
- Dynamic answers can only be generated if there is at least one value.
```

doc/decisions/0012-repeating-answer-options.md
```diff
@@ -0,0 +1,274 @@
...
+          },
+          "decimal_places": 2
+        }
+      ]
+    }
+  }
+}
+```
+
+- `list_item_id` can now be an answer value or the output of a function, not just list collector ids.
```
MebinAbraham, line 140: Mention the remove of spaces from values for it can be used as an id for storage.

doc/decisions/0012-repeating-answer-options.md
```diff
@@ -0,0 +1,274 @@
...
+},
+{
+  "answer_id": "percentage-of-shopping"
+  "value": 30
+  "list_item_id": fdhjfl
+}
+```
+
+For other sources, we would set the `list_item_id` and use it the same way it is used for list collector sources but instead of it being set to an automatically generated value we would set it to to the option value with any spaces replaced with hyphens as per the checkbox example below:
+```
```
MebinAbraham, line 240: use json or py syntax highlighting.

doc/decisions/0012-repeating-answer-options.md
```diff
@@ -0,0 +1,274 @@
...
+  "list_item_id": "Aldi"
+}
+```
+
+This is something we will need to discuss with downstream teams as they will need to know we are using it in this way. We will also need to discuss it with Author because they will need to know that they need to be set but must be unique within the `dynamic_answers` object. These will be then concatenated with the `answer_id` and any spaces will be replaced with dashes in order to be valid ids when set in the HTML.
+
+## Answer Codes
+
+As answer ids used for dynamic answers will only be prefixed values in the schema, only one answer code would be able to be set against each dynamic answer (using the id prefix) e.g.
+```
```
MebinAbraham, line 263: Same as above

MebinAbraham (approved):

Yuyuutsu (approved):

creativeappdevamit (approved):

PR #181, opened 2023-03-08
Chaning regex to accommodate upper case metadata identifiers for PHM

creativeappdevamit:
Current metadata identifiers regex doesn't allow upper case, although the current version of AJV was unable to detect this and provides false positive, with the new AJV version update this scenario will fail for PHM as PHM schema was designed with upper case metadata identifiers. Hence changing the regex to accommodate upper case. 

### How to review
Check the regex still matches the basic validation including upper case.

### Checklist


creativeappdevamit:
[Update questionnaire validator to latest AJV and JSON draft. (3 SP)](https://trello.com/c/KMuEMGUd/5512-update-questionnaire-validator-to-latest-ajv-and-json-draft-3-sp)

Yuyuutsu (approved):

PR #187, opened 2023-03-24
Add dynamic answers list validation

petechd:
Adds validation of the new dynamic answers object as proposed in https://github.com/ONSdigital/eq-questionnaire-validator/blob/main/doc/decisions/0012-repeating-answer-options.md.

### How to review
Validate new dynamic answers schemas.

### Checklist


petechd:
[Iteration 3 - Dynamically creating answers based on a list (8 SP)](https://trello.com/c/eO99WHvl/5323-iteration-3-dynamically-creating-answers-based-on-a-list-8-sp)

schemas/questions/types/general.json
```diff
@@ -35,9 +35,75 @@
...
+      "dynamic_answers": {
+        "type": "object",
+        "properties": {
+          "values": {
+            "type": "object",
+            "source": ["list"],
+            "identifier": {
+              "$ref": "../../common_definitions.json#/non_empty_string"
+            }
+          }
```
MebinAbraham, line 47: This should reference the existing list source.

schemas/questions/types/general.json
```diff
@@ -35,9 +35,75 @@
...
+        },
+        "answers": {
+          "type": "array",
+          "items": {
+            "oneOf": [
+              {
+                "$ref": "../../answers/address.json#/answer"
```
MebinAbraham, line 54: Address and Relationship type can be removed.

schemas/questions/types/general.json
```diff
@@ -35,9 +35,75 @@
...
+              {
+                "$ref": "../../answers/mobile_number.json#/answer"
+              }
+            ]
+          }
+        }
+      },
       "answers": {
         "type": "array",
-        "minItems": 1,
```
MebinAbraham, line 40: This shouldn't have changed. If we have answers, then it should be min 1.

You need to update the JSON validation to be conditional answer if dynamic answers is defined. One of them must be defined.

schemas/questions/types/general.json
```diff
@@ -35,9 +35,75 @@
         "type": "string",
         "enum": ["General"]
       },
+      "dynamic_answers": {
+        "type": "object",
+        "properties": {
```
MebinAbraham, line 40: Need to make `values` and `answers` a mandatory property and need to disable additional properties.

MebinAbraham (commented):
Answer labels currently don't support placeholder, that will need to be changed here, as dynamic answers needs to reference the item in the label.

schemas/questions/types/general.json
```diff
@@ -35,9 +35,75 @@
...
+          "values": {
+            "type": "object",
+            "source": ["list"],
+            "identifier": {
+              "$ref": "../../common_definitions.json#/non_empty_string"
+            }
+          }
+        },
+        "answers": {
```
MebinAbraham, line 49: We need to write some custom validation to validate answer source used inside "dynamic_answers.answers", either in placeholders or validation. Since answer sources by default uses the list item id of the item, any answer sources that references non-repeating answers will need a "list_item_selector" defined otherwise it will error. This needs to be validate and also tested on the runner side.

petechd: This one is outdated as validation is not needed anymore since we made the relevant changes in runner.

schemas/questions/types/general.json
MebinAbraham: Still outstanding.

Mine items can stay where it is.

You need to make add JSON logic to check at least either dynamic_answers or answers are defined.

schemas/questions/types/general.json
```diff
@@ -38,9 +38,62 @@
...
+        "properties": {
+          "values": {
+            "oneOf": [
+              {
+                "$ref": "../../value_sources.json#/list_value_source"
+              }
+            ]
+          },
+          "answers": {
+            "type": "array",
```
MebinAbraham, line 52: Should also have min items here.

schemas/questions/types/general.json
petechd: Added this logic now, added back min items.

MebinAbraham (commented):
There are many methods that reference `"answer"` which needs to be updated to handle it not existing and also need to handle making use of `dynamic_answers`. For example, things to get all answer ids, doesn't get the dynamic answer ids as it hasn't been updated. New units tests should be added accordingly.

schemas/questions/types/general.json
```diff
@@ -95,7 +150,17 @@
...
+      "not": {
+        "required": ["dynamic_answers"]
+      }
+    },
+    "then": {
+      "required": ["id", "type", "title", "answers"]
+    },
+    "else": {
+      "required": ["id", "type", "title"]
+    },
```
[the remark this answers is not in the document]
petechd: Had to add both valid and invalid. Valid mainly for unit test purposes.

tests/test_questionnaire_schema.py
```diff
@@ -215,3 +215,103 @@ def test_get_block_id_by_answer_id():
...
+
+
+def test_dynamic_answers():
+    filename = "schemas/valid/test_dynamic_answers_list_source.json"
+
+    questionnaire_schema = QuestionnaireSchema(_open_and_load_schema_file(filename))
+
+    answers = list(questionnaire_schema.answers)
+
+    assert answers[7] == {
```
katie-gardner, line 258: same thought as above

katie-gardner (changes requested):
Happy with functionality of the changes. Altering schemas in a way that should cause test failures does, and valid schemas are passing, changes make sense and cover a good range of scenarios. Just have a few minor comments and questions

schemas/questions/types/general.json
```diff
@@ -38,6 +38,61 @@
...
+          },
+          "answers": {
+            "type": "array",
+            "minItems": 1,
+            "items": {
+              "oneOf": [
+                {
+                  "$ref": "../../answers/currency.json#/answer"
+                },
+                {
```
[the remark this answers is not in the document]
petechd: There was a list of answers out of scope for this looping iteration, address is one of them. I can't remember if it's on the card somewhere but check this comment -> https://github.com/ONSdigital/eq-questionnaire-validator/pull/187#discussion_r1163352890.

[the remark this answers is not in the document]
katie-gardner: ahh I missed that - thanks

app/validators/questionnaire_schema.py
```diff
@@ -335,13 +348,22 @@ def get_list_collector_answer_ids(self, block_id):
     @lru_cache
     def get_all_answer_ids(self, block_id):
         questions = self.get_all_questions_for_block(self.blocks_by_id[block_id])
-        return {
-            answer["id"] for question in questions for answer in question["answers"]
-        }
+        answer_ids = []
+        for question in questions:
+            if question.get("answers", []):
```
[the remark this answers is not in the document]
petechd: Added, thanks.

app/validators/questionnaire_schema.py
```diff
@@ -351,13 +373,25 @@ def _get_path_id(self, path):
     @lru_cache
     def get_block_id_by_answer_id(self, answer_id):
         for question, context in self.questions_with_context:
-            for answer in question.get("answers", []):
-                if answer_id == answer["id"]:
+            if question.get("dynamic_answers", {}):
```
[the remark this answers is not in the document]
petechd: Added, thanks.

tests/test_questionnaire_schema.py
```diff
@@ -215,3 +215,103 @@ def test_get_block_id_by_answer_id():
...
 
     assert block_id == "confirmation-1"
+
+
+def test_dynamic_answers_with_context():
+    filename = "schemas/valid/test_dynamic_answers_list_source.json"
+
+    questionnaire_schema = QuestionnaireSchema(_open_and_load_schema_file(filename))
+
+    assert questionnaire_schema.answers_with_context["percentage-of-shopping"] == {
```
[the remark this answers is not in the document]
petechd: I think the more we test, the better. Just shows unit testing in validator has huge gaps and I am happy to add non dynamic to the tests as well. Same regarding the comment below.

app/validators/questionnaire_schema.py
```diff
@@ -335,13 +348,22 @@ def get_list_collector_answer_ids(self, block_id):
     @lru_cache
     def get_all_answer_ids(self, block_id):
         questions = self.get_all_questions_for_block(self.blocks_by_id[block_id])
-        return {
-            answer["id"] for question in questions for answer in question["answers"]
-        }
+        answer_ids = []
```
kylelawsonAND, line 351: Since this is converted to a set before returning and only added to, couldn't this be initialised as a set, avoiding the conversion at the end?

kylelawsonAND (changes requested):
Just one comment - tests all working and makes sense 👍

app/validators/questionnaire_schema.py
petechd: For some reason I thought since sets are immutable it is better to add to a list and then convert to set. Removed now, thanks!

kylelawsonAND (approved):

app/validators/questionnaire_schema.py
```diff
@@ -351,13 +373,25 @@ def _get_path_id(self, path):
...
+            if dynamic_answers := question.get("dynamic_answers", {}):
+                if block_id := self.capture_block_id_for_answer(
+                    answer_id, context, dynamic_answers["answers"]
+                ):
+                    return block_id
+            if answers := question.get("answers", []):
+                if block_id := self.capture_block_id_for_answer(
+                    answer_id, context, answers
+                ):
+                    return block_id
```
MebinAbraham, line 385: Based on previous comments, this can be simplified to:
```suggestion
            if block_id := self.capture_block_id_for_answer(
                answer_id, context, self.get_answers_from_question(question)
            ):
                return block_id
```

app/validators/questions/question_validator.py
```diff
@@ -10,7 +10,11 @@ class QuestionValidator(Validator):
     def __init__(self, schema_element, schema=None):
         super().__init__(schema_element)
         self.question = schema_element
-        self.answers = self.question.get("answers", [])
+        if self.question.get("dynamic_answers", {}):
+            self.dynamic_answers = self.question["dynamic_answers"]["answers"]
+        else:
+            self.dynamic_answers = []
+        self.answers = self.question.get("answers", []) + self.dynamic_answers
```
MebinAbraham, line 17: Can be simplified to:
```suggestion
        self.answers = QuestionnaireSchema.get_answers_from_question(self.question)
```

Note you need to import `QuestionnaireSchema`

MebinAbraham, line 17: Can be simplified to:
```suggestion
        self.answers = QuestionnaireSchema.get_answers_from_question(self.question)
```

Note you need to import `QuestionnaireSchema`

app/validators/sections/section_validator.py
```diff
@@ -116,23 +116,9 @@ def validate_question(self, block_or_variant):
...
-                    "Number",
-                ]:
-                    self.add_error(
-                        error_messages.UNSUPPORTED_QUESTION_SUMMARY_ANSWER_TYPE,
-                        answer_id=answer["id"],
-                    )
-                self.errors += answer_validator.errors
+            self._validate_answers(question, question.get("answers", []))
+            if dynamic_answers := question.get("dynamic_answers", {}):
+                self._validate_answers(question, dynamic_answers["answers"])
```
MebinAbraham, line 121: Can be inline, just use `get_answers_from_question`

MebinAbraham (commented):
A lot of the method added become redundant if you add a helper method to get answers. At the moment, you are checked for normal and dynamic in a lot of places which should be avoided.

Other than that, looks good.

app/validators/questions/question_validator.py
petechd: Had to do this as a pushed commit as we need the import.

petechd: Duplicate below?

app/validators/questionnaire_schema.py
petechd: Had to do this by hand because the method name has changed.

petechd:
Should we fix overriding get default in?:
```python
  for answer in question.get("answers", []):
      for option in answer.get("options", []):
          action = option.get("action")
```
the options one was done in the past..

MebinAbraham:
> Should we fix overriding get default in?:
> 
> ```python
>   for answer in question.get("answers", []):
>       for option in answer.get("options", []):
>           action = option.get("action")
> ```
> 
> the options one was done in the past..

No, this is valid, because it is a for loop.

app/validators/questionnaire_schema.py
```diff
@@ -351,13 +360,22 @@ def _get_path_id(self, path):
...
-                if answer_id == answer["id"]:
+            if block_id := self.return_block_id_for_answer(
+                answer_id=answer_id,
+                answers=self.get_answers_from_question(question),
+                context=context,
+            ):
+                return block_id
+
+    @staticmethod
+    def return_block_id_for_answer(*, answer_id, answers, context):
```
MebinAbraham, line 371: Would be more appropriate to call it `get_block_id_for_answer`

app/validators/sections/section_validator.py
```diff
@@ -116,23 +119,9 @@ def validate_question(self, block_or_variant):
...
-                    "Checkbox",
-                    "Number",
-                ]:
-                    self.add_error(
-                        error_messages.UNSUPPORTED_QUESTION_SUMMARY_ANSWER_TYPE,
-                        answer_id=answer["id"],
-                    )
-                self.errors += answer_validator.errors
+            self._validate_answers(
+                question, QuestionnaireSchema.get_answers_from_question(question)
```
MebinAbraham, line 123: Do this inline, call `QuestionnaireSchema.get_answers_from_question` inside _validate_answer` since you are already passing question.

schemas/questions/types/general.json
```diff
@@ -38,6 +38,61 @@
...
+      "dynamic_answers": {
+        "type": "object",
+        "properties": {
+          "values": {
+            "oneOf": [
+              {
+                "$ref": "../../value_sources.json#/list_value_source"
+              }
+            ]
+          },
```
MebinAbraham, line 50: Can be inline:
```suggestion
          "values": {
            "$ref": "../../value_sources.json#/list_value_source"
          },
```

petechd:
> 
> No, this is valid, because it is a for loop.

I get it now, thanks.

tests/schemas/valid/test_dynamic_answers_list_source.json
```diff
@@ -0,0 +1,322 @@
...
+                  ]
+                }
+              },
+              "question": {
+                "dynamic_answers": {
+                  "values": {
+                    "source": "list",
+                    "identifier": "supermarkets"
+                  },
+                  "answers": [
```
MebinAbraham, line 288: We don't have an example where dynamic answers have more than 1 answer. We should add another answer of a different type, i.e Numbers. Same for runner.

petechd: Added, thanks.

MebinAbraham (approved):

PR #189, opened 2023-04-04
Update validator for only one question definition

rmccar:
Updates validator to fit with the runner change to only allow one question definition

Also fixes a typo

### How to review
- Tests pass
- Changes make sense

### Checklist


MebinAbraham:
This is a breaking change and we can't get this merged until Author can support this. The eq-questionnaire-schemas repo will need to be updated to fix existing schemas. Realistic we can't merge this without Author being on board otherwise they will have to keep hand cranking the change.

rmccar:
> This is a breaking change and we can't get this merged until Author can support this. The eq-questionnaire-schemas repo will need to be updated to fix existing schemas. Realistically we can't merge this without Author being on board otherwise they will have to keep hand cranking the change.
> 
> EDIT: Have spoken to Martyn and they can accommodate this fairly quick. So this time around we may be okay, in future we need to keep this in mind as we won't be able to just make breaking changes especially with the CIR in place. However, to make this work, you will need to convert all the current schemas in eq-questionnaire-schemas to use the new format, update eq-translations as well. This also has knock on effect on CRIS as translations will need redoing, it is going live shortly, so this might not be a good time to do this. We'll have a catch-up tomorrow to discuss.

As discussed have made the change to runner to allow both object type `definition` and array type `definitions` for the time being so is no longer a breaking change

schemas/common_definitions.json
```diff
@@ -18,6 +18,14 @@
...
     "minItems": 1,
+    "maxItems": 1,
+    "items": {
+      "$ref": "#/definition"
+    }
+  },
+  "question_definition": {
+    "description": "Allows customisation of question definition title and description.",
+    "type": "object",
     "items": {
```
MebinAbraham, line 29: "items" is for an array. Is this even needed?. You can just use `definition` directly?

MebinAbraham:
eq-translations needs to be updated to support the new definitions FYI.

Yuyuutsu (approved):
LGTM!

MebinAbraham (approved):

PR #191, opened 2023-04-14
Update max_number and min_number values

Yuyuutsu:
Update max and min number values for validator. There should be 14 negative digits (12 whole numbers and 2 decimal numbers) and 15 positive digits (13 whole numbers and 2 decimal numbers).

There is a test needed for `validate_value_in_limits` But will be added in another Pull Request.
UPDATE: 
- Now also checks any reference values using answer sources (number validation)
- Made `def _get_numeric_value_for_value_source` into a external function.
### How to review

- Run functional tests
- Check if the changes made fit the [trello card](https://trello.com/c/0yJIjinU/5658-enable-up-to-15-digits-in-relevant-fields-1-sp)
- Run validator against a instance of runner in the `update-max-length-to-15-digits` branch
- All relevant tests are edited / made? 
### Checklist


Yuyuutsu:
[Enable up to 15 digits in relevant fields (1 SP)](https://trello.com/c/0yJIjinU/5658-enable-up-to-15-digits-in-relevant-fields-1-sp)

Yuyuutsu:
[Enable up to 15 digits in relevant fields (1 SP)](https://trello.com/c/0yJIjinU/5658-enable-up-to-15-digits-in-relevant-fields-1-sp)

tests/validators/answers/test_number_answer_validator.py
```diff
@@ -9,8 +9,8 @@ def test_minimum_value():
         "id": "answer-4",
         "label": "Max/Min out of system limits",
         "mandatory": False,
-        "maximum": {"value": 99999999999},
-        "minimum": {"value": -99999999999},
+        "maximum": {"value": 99999999999999},
```
petechd, line 12: Can you use the `_` separator here and below as well? It is hard to read.

Yuyuutsu: Added 👍

app/validators/answers/number_answer_validator.py
```diff
@@ -1,7 +1,7 @@
 from app.validators.answers.answer_validator import AnswerValidator
 
-MAX_NUMBER = 9999999999
-MIN_NUMBER = -999999999
+MAX_NUMBER = 9_999_999_999_999
+MIN_NUMBER = -999_999_999_999
```
MebinAbraham, line 4: This is only max 13 digits. Card say max 15 for positive, 14 for negative.

Yuyuutsu: My bad it was my assumption, since the card said we already support 12 digits with only 10 9's in the max_number variable

app/validators/answers/number_answer_validator.py
```diff
@@ -71,6 +73,42 @@ def validate_value_in_limits(self):
...
+                and answer_source_max_value > MAX_NUMBER
+            ):
+                self.add_error(
+                    self.MAXIMUM_GREATER_THAN_LIMIT,
+                    value=answer_source_max_value,
+                    limit=MAX_NUMBER,
+                )
+
+        if isinstance(answer_source_min_value, dict):
+            source = answer_source_min_value.get("source")
```
Yuyuutsu, line 95: Not sure to make this a function since it is repeated?

app/validators/answers/number_answer_validator.py
```diff
@@ -71,6 +73,42 @@ def validate_value_in_limits(self):
...
+            answers_with_context = self.questionnaire_schema.numeric_answer_ranges
+            answer_source_min_value = (
+                self.questionnaire_schema.get_numeric_value_for_value_source(
+                    source, answer_source_min_value, answers_with_context
+                )
+            )
+            if (
+                isinstance(answer_source_min_value, int)
+                and answer_source_min_value > MIN_NUMBER
+            ):
```
MebinAbraham, line 105: `answer_source_min_value` is an optional dict, it wont be an int. Need to add tests to value source which would demo this not working.

This is also reassigning `answer_source_min_value`, at one point it is a value source, at line 97 it because the "referred answer" which is an optional dict, use diff variable names for clarify.

app/validators/answers/number_answer_validator.py
```diff
@@ -71,6 +73,42 @@ def validate_value_in_limits(self):
...
+            source = answer_source_min_value.get("source")
+            answers_with_context = self.questionnaire_schema.numeric_answer_ranges
+            answer_source_min_value = (
+                self.questionnaire_schema.get_numeric_value_for_value_source(
+                    source, answer_source_min_value, answers_with_context
+                )
+            )
+            if (
+                isinstance(answer_source_min_value, int)
+                and answer_source_min_value > MIN_NUMBER
```
MebinAbraham, line 104: With negative numbers it should be `min < MIN_NUMBER`. `min > MIN_NUMBER` would mean if `-10` was the limit, `-11` would pass as `-11` is not greater than `-10`.

app/validators/answers/number_answer_validator.py
```diff
@@ -60,6 +60,8 @@ def validate_mandatory_has_no_default(self):
     def validate_value_in_limits(self):
```
MebinAbraham, line 60: Now that this uses the schema, this should be called after `validate_referred_numeric_answer_decimals` in the main `validate()` method.

app/validators/answers/number_answer_validator.py
Yuyuutsu: Forgot to swap it when copying and pasting, updated 👍

app/validators/answers/number_answer_validator.py
kylelawsonAND: Think it's fine as there's an operator change between these blocks, not just variable changes, so doesn't feel worth the complication to make a function in this scenario. 
Rule of 3 - if you have to repeat it three times, make a function, but don't use as an excuse not to be DRY - Don't Repeat Yourself. In this case it feels alright to keep these repeated parts.

app/validators/answers/number_answer_validator.py
Yuyuutsu: Updated 👍

app/validators/answers/number_answer_validator.py
Yuyuutsu: Ahh I see, thanks thats a useful mnemonic - I will be using that from now on

app/validators/answers/number_answer_validator.py
```diff
@@ -71,6 +74,42 @@ def validate_value_in_limits(self):
...
+            source = answer_source_min_value.get("source")
+            answers_with_context = self.questionnaire_schema.numeric_answer_ranges
+            value_source_min_value = (
+                self.questionnaire_schema.get_numeric_value_for_value_source(
+                    source, answer_source_min_value, answers_with_context
+                )
+            )
+            if (
+                isinstance(value_source_min_value, int)
+                and value_source_min_value > MIN_NUMBER
```
rmccar, line 105:
```suggestion
                and value_source_min_value > MIN_NUMBER
```
Doesn't this logic need to be reversed? We want to add an error if the value is less than the minimum limit

app/validators/answers/number_answer_validator.py
```diff
@@ -71,6 +74,42 @@ def validate_value_in_limits(self):
...
+            source = answer_source_max_value.get("source")
+            answers_with_context = self.questionnaire_schema.numeric_answer_ranges
+            value_source_max_value = (
+                self.questionnaire_schema.get_numeric_value_for_value_source(
+                    source, answer_source_max_value, answers_with_context
+                )
+            )
+            if (
+                isinstance(value_source_max_value, int)
+                and value_source_max_value < MAX_NUMBER
```
rmccar, line 87:
```suggestion
                and value_source_max_value > MAX_NUMBER
```
Same here we want to add an error if the value is greater than the max limit

rmccar (commented):
Other than these looks fine

app/validators/answers/number_answer_validator.py
```diff
@@ -60,6 +61,8 @@ def validate_mandatory_has_no_default(self):
     def validate_value_in_limits(self):
```
MebinAbraham, line 61: This can be simplified/fixed to something like:

```py
    def validate_value_in_limits(self):
        min_value = self.answer.get("minimum", {}).get("value", 0)
        max_value = self.answer.get("maximum", {}).get("value", 0)
        answer_ranges = self.questionnaire_schema.numeric_answer_ranges

        if isinstance(min_value, int) and min_value < MIN_NUMBER:
            self.add_error(
                self.MINIMUM_LESS_THAN_LIMIT, value=min_value, limit=MIN_NUMBER
            )
        elif isinstance(min_value, dict):
            source = min_value["source"]
            referred_answer = (
                self.questionnaire_schema.get_numeric_value_for_value_source(
                    source, min_value, answer_ranges
                )
            )
            if referred_answer["min"] < MIN_NUMBER:
                self.add_error(
                    self.MINIMUM_LESS_THAN_LIMIT, value=referred_answer["min"], limit=MIN_NUMBER
                )

        if isinstance(max_value, int) and max_value > MAX_NUMBER:
            self.add_error(
                self.MAXIMUM_GREATER_THAN_LIMIT, value=max_value, limit=MAX_NUMBER
            )
        elif isinstance(min_value, dict):
            source = max_value["source"]
            referred_answer = (
                self.questionnaire_schema.get_numeric_value_for_value_source(
                    source, min_value, answer_ranges
                )
            )
            if referred_answer["max"] > MAX_NUMBER:
                self.add_error(
                    self.MAXIMUM_GREATER_THAN_LIMIT, value=referred_answer["max"], limit=MAX_NUMBER
                )
```

You will need to add tests for value source checks

tests/validators/answers/test_number_answer_validator.py
```diff
@@ -9,8 +9,8 @@ def test_minimum_value():
         "id": "answer-4",
         "label": "Max/Min out of system limits",
         "mandatory": False,
-        "maximum": {"value": 99999999999},
-        "minimum": {"value": -99999999999},
+        "maximum": {"value": 9_999_999_999_999_999},
+        "minimum": {"value": -999_999_999_999_999},
```
MebinAbraham, line 13: While you are here, update this to use the `MIN/MAX_NUMBER` constants.

app/validators/answers/number_answer_validator.py
```diff
@@ -1,7 +1,7 @@
 from app.validators.answers.answer_validator import AnswerValidator
 
-MAX_NUMBER = 9999999999
-MIN_NUMBER = -999999999
+MAX_NUMBER = 999_999_999_999_999
+MIN_NUMBER = -99_999_999_999_999
```
kylelawsonAND, line 4: If this goes ahead before Donna gets an answer - remember to change min to 15 9s. Feel free to resolve this when you change it for merging or if the decision is made to stay 14 👍

kylelawsonAND (changes requested):

app/validators/answers/number_answer_validator.py
Yuyuutsu: Cheers ✌️

app/validators/answers/number_answer_validator.py
```diff
@@ -65,11 +66,39 @@ def validate_value_in_limits(self):
...
+        elif isinstance(min_value, dict):
+            source = min_value["source"]
+            answer_ranges = self.questionnaire_schema.numeric_answer_ranges
+            referred_answer = (
+                self.questionnaire_schema.get_numeric_value_for_value_source(
+                    source, min_value, answer_ranges
+                )
+            )
+            if referred_answer["max"] > MAX_NUMBER:
+                self.add_error(
```
katie-gardner, line 97: I think you meant for this to be max_value, currently its min_value twice

katie-gardner (changes requested):

app/validators/answers/number_answer_validator.py
Yuyuutsu: Changed max_value 👍 on line 99

app/validators/answers/number_answer_validator.py
Yuyuutsu: UPDATE TO REVIEWERS: After discussion I am going to add tests in another Pull request later when I have time .

MebinAbraham (approved):
Expediting this. Test to be added in a separate PR, functionally working as expected so ✅

kylelawsonAND (approved):
All seems fine - resolve my existing comment if it is being left as is 👍

PR #193, opened 2023-05-04
Update definition definitions

rmccar:
This updates the definitions to remove the items object which is causing the definitions to not be validated properly. Currently `test_question_definition` is passing validation when it should fail. This is because the definition is using "content" instead of "contents".

### How to review
- Test that using the Runner main branch and validator main branch passes validation
- Test that this branch along with the Runner main branch fails validation on `test_question_definition`
- Test that this branch along with the Runner `update-question-definition` branch where "content" is renamed to "contents" and is an array of objects (which is what it should be) in the `test_question_definition` it passes validation 

### Checklist

### DIFF
--- a/schemas/common_definitions.json
+++ b/schemas/common_definitions.json
@@ -76,19 +76,17 @@
   },
   "definition": {
     "description": "Definition title and contents.",
-    "items": {
-      "type": "object",
-      "required": ["title", "contents"],
-      "properties": {
-        "title": {
-          "$ref": "#/non_empty_string"
-        },
-        "contents": {
-          "$ref": "#/contents"
-        }
+    "type": "object",
+    "required": ["title", "contents"],
+    "properties": {
+      "title": {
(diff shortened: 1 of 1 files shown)


rmccar:
[Rework the onsQuestion component to allow runner to use it more efficiently (3 SP)](https://trello.com/c/9yN6rAG9/5625-rework-the-onsquestion-component-to-allow-runner-to-use-it-more-efficiently-3-sp)

katie-gardner (approved):
Tests pass on main branches, fail when only the validator branch is checked out, and pass again with both updates. Changes to schema make sense. 👍

kylelawsonAND (approved):
Both test_benchmark_business and test_question_definition fail when runner main uses new validator, then passes when both runner and validator are new branch ✅

kylelawsonAND (approved):
Both test_benchmark_business and test_question_definition fail when runner main uses new validator, then passes when both runner and validator are new branch ✅

PR #196, opened 2023-05-11
Design proposal for referencing supplementary data

MebinAbraham:
Proposal to define how eQ Runner will reference data from SDS.

### How to review
- Ensure the documented decision is appropriate.
- Has anything been missed?.

### Checklist


doc/decisions/0013-referencing-supplementary-data.md
```diff
@@ -0,0 +1,160 @@
...
+
+When referencing data for a repeating item, there will be no need to use the keys such as `"items"` in the source definition. This structure provides a generic way to define repeating items and for Runner to parse it with enough context.
+
+Referencing data for repeating items works similarly to non-repeating items. The reference will be made using the list item name.
+
+A key point to understand here is that referencing data for a repeating item will only be supported in the context of a "repeat". For example, inside the list collector child pages, such as the repeating questions, or anywhere inside a repeating section.
+This is because, in both cases, Runner will have access to the list item id of the current item, which is what Runner will use to look up the data for the correct item. This is identical to how values for **answers** are resolved inside a repeat. We use the answer ID and the list item ID to resolve the value.
+
+For both the examples below, we'll assume the list item id for the location (URL) is `abCdE`, which behind the scenes is Runner's ID which maps to the SDS ID `89929001` (`items.products.values.identifier`).
```
kylelawsonAND, line 130: Could we elaborate on how the Runner Item ID maps to SDS Item ID when referencing  value for an SDS list item? 
Above it's mentioned that EQ provides the Runner Item ID to SDS Item ID to downstream when the list item is created and submitted, how will a later survey have context of this mapping that happened in a prev survey?

kylelawsonAND (changes requested):

doc/decisions/0013-referencing-supplementary-data.md
kylelawsonAND: If this won't be known until implementation, that's fair, can then be left as is

katie-gardner (approved):
Aligns with schema definitions (subject to key field changing) and well explained

doc/decisions/0013-referencing-supplementary-data.md
MebinAbraham: The runner ID won't be used by later surveys, it is purely for runner and for downstream to work out the instance of an answer. It tells them which answers belong to which item. Since for prepop list, it will still use list item ids, the mapping we send downstream is just so they can do a mapping to say so "answer-1" "instance-1" is for product id `89929001` etc.

> how the Runner Item ID maps to SDS Item ID when referencing value for an SDS list item?

Do you mean literally how it will map?. The sds item id won't be used for logic until it is needed for the downstream mapping, all runner lookups will be by runner item id.

kylelawsonAND: I'm with you now, sounds good 👍

berroar (approved):

petechd (approved):

rmccar (approved):

kylelawsonAND (approved):
Update to reflect usage of "identifier" looks good

Yuyuutsu (approved):
Had a read through, proposal seems good to me 👍

petechd (approved):

katie-gardner (approved):
Newest changes for identifier all look good 👍

martyncolmer (approved):
I'm happy with this

PR #197, opened 2023-05-16
Grand calculated summary validation

katie-gardner:
This introduces a new block type for grand calculated summaries, which is similar to the calculated summary block, except that it only supports calculations of the new style and it only supports value sources of `calculated_summary`. This is the validation component of the [Feature PR on Runner](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1107)

### How to review
Test that validator passes when using this branch alongside the grand calculated summary PR from runner. Check that the schema is correct and that any invalid cases on the scenarios for the ticket have been covered. Check that this is inline with the validation for calculated summaries.

### Checklist
* [ ] Validation passes on runners main branch
* [ ] Validation passes on the grand calculated summary PR


katie-gardner:
[Grand Calculated Summary: Initial feature build (5 SP)](https://trello.com/c/mcqdwMMQ/5062-grand-calculated-summary-initial-feature-build-5-sp)

app/validators/blocks/calculation_block_validator.py
```diff
@@ -0,0 +1,52 @@
+from app.validators.blocks.block_validator import BlockValidator
+
+
+def key_value_is_unique(dictionaries: list[dict], key: str) -> bool:
```
MebinAbraham, line 4: Might be clearer as `is_value_for_key_unique`?

app/validators/blocks/grand_calculated_summary_block_validator.py
```diff
@@ -0,0 +1,100 @@
+from app.validators.blocks.calculation_block_validator import CalculationBlockValidator
+from app.validators.questionnaire_schema import (
+    find_dictionary_duplicates,
+    find_duplicates,
+)
+
+
+class GrandCalculatedSummaryBlockValidator(CalculationBlockValidator):
```
MebinAbraham, line 8: This is currently not being used for real validation. You need to update the factory method mapping https://github.com/ONSdigital/eq-questionnaire-validator/blob/main/app/validators/blocks/__init__.py#L17-L25

app/validators/blocks/grand_calculated_summary_block_validator.py
```diff
@@ -0,0 +1,100 @@
...
+    CALCULATED_SUMMARIES_HAS_DUPLICATES = (
+        "Duplicate calculated summaries in block's value sources"
+    )
+    CALCULATED_SUMMARY_WITH_DUPLICATE_ANSWERS = "Cannot have two calculated summaries referencing exactly the same answers in a grand calculated summary"
+    CALCULATED_SUMMARY_AFTER_GRAND_CALCULATED_SUMMARY = "Cannot have a grand calculated summary before a calculated summary that it depends on"
+    GRAND_CALCULATED_SUMMARY_HAS_INVALID_SOURCE = (
+        "All value sources for grand calculated summary must be calculated summaries"
+    )
```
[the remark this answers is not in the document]
katie-gardner: Good point - deleted

app/validators/blocks/grand_calculated_summary_block_validator.py
```diff
@@ -0,0 +1,100 @@
...
+    find_dictionary_duplicates,
+    find_duplicates,
+)
+
+
+class GrandCalculatedSummaryBlockValidator(CalculationBlockValidator):
+    CALCULATED_SUMMARIES_HAS_DUPLICATES = (
+        "Duplicate calculated summaries in block's value sources"
+    )
+    CALCULATED_SUMMARY_WITH_DUPLICATE_ANSWERS = "Cannot have two calculated summaries referencing exactly the same answers in a grand calculated summary"
```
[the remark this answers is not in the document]
katie-gardner: Done 👍

app/validators/blocks/grand_calculated_summary_block_validator.py
```diff
@@ -0,0 +1,100 @@
...
+from app.validators.questionnaire_schema import (
+    find_dictionary_duplicates,
+    find_duplicates,
+)
+
+
+class GrandCalculatedSummaryBlockValidator(CalculationBlockValidator):
+    CALCULATED_SUMMARIES_HAS_DUPLICATES = (
+        "Duplicate calculated summaries in block's value sources"
+    )
```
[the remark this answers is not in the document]
katie-gardner: Both this and calculated summary were missing `unique_items` so I've added that and removed the checks we don't need, thanks

app/validators/blocks/calculation_block_validator.py
katie-gardner: Updated 👍

app/validators/blocks/grand_calculated_summary_block_validator.py
katie-gardner: Thank you - fixed 👍

MebinAbraham (approved):

app/validators/questionnaire_schema.py
```diff
@@ -442,7 +454,11 @@ def _get_numeric_value(self, defined_value, system_default, answer_ranges):
         return system_default
 
     @staticmethod
-    def get_calculated_answer_ids(block):
+    def get_calculation_block_ids(block, source_type: str):
```
petechd, line 457: Not a big problem but shouldn't we provide both type hints here?

katie-gardner: the majority of the file isn't type hinted so I wasn't sure how much value it would add but no harm in adding them - updated 👍

MebinAbraham:
It doesn't look like we have validation to prevent a calculated summary from a repeating section being used n a GCS outside a repeat. However, given where this is at, I think we get this in as is. I've added that validation to the card that will add support for GCS in a repeating section.

app/validators/blocks/calculated_summary_block_validator.py
```diff
@@ -1,59 +1,29 @@
...
         "Answer ids for calculated summary must be set before calculated summary block"
     )
     ANSWER_SET_IN_DIFFERENT_SECTION_FOR_CALCULATED_SUMMARY = "Answer ids for calculated summary must be set in the same section as the calculated summary block"
 
     def __init__(self, block, questionnaire_schema):
         super().__init__(block, questionnaire_schema)
-        self.answers_to_calculate = self.questionnaire_schema.get_calculated_answer_ids(
-            self.block
+        self.answers_to_calculate = self.questionnaire_schema.get_calculation_block_ids(
+            self.block, "answers"
```
kylelawsonAND, line 13: Its only two params but I'm enjoying the push to using kwargs when raw values are passed as args

schemas/blocks/grand_calculated_summary.json
```diff
@@ -0,0 +1,43 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "block": {
+    "type": "object",
+    "properties": {
+      "id": {
+        "$ref": "../common_definitions.json#/identifier"
+      },
+      "type": {
+        "type": "string",
+        "enum": ["GrandCalculatedSummary"]
```
kylelawsonAND, line 12: I've done the same thing in my validator changes - using a single value array for enum - just wondering is this the done thing, should we be putting only a single value, would that work?

katie-gardner: Yeah its been done like this throughout and looks like its the standard for ajv https://ajv.js.org/json-type-definition.html#enum-form so I think thats what it is expecting even if usually just the one type

app/validators/blocks/calculated_summary_block_validator.py
katie-gardner: Nah happy with that suggestion, good idea 👍  
done

kylelawsonAND (approved):

berroar (approved):
LGTM 👍

petechd (approved):

PR #198, opened 2023-05-16
Update translations version to 4.9.0

rmccar:
Small PR to update eq-translations version to the latest released version of eq-translations (v4.9.0).

### How to review 
Check the change is to the correct version.

### Checklist


katie-gardner (changes requested):

poetry.lock
```diff
@@ -1,19 +1,14 @@
-# This file is automatically @generated by Poetry and should not be changed by hand.
-
 [[package]]
 name = "astroid"
-version = "2.12.13"
+version = "2.15.5"
```
katie-gardner, line 3: this looks like its updated a bunch of other libraries as a byproduct

katie-gardner (changes requested):

poetry.lock
katie-gardner: I think the update command needs to specify eq-translations only

rmccar: When I recreated the poetry file using `poetry update eq-translations` it automatically updated the other ones

rmccar:
[Update eq-translations version](https://trello.com/c/UTgkU9qU/5714-update-eq-translations-version)

poetry.lock
katie-gardner: Yeah it does look that way, or that if you do want to it would have to be specified in the toml file https://python-poetry.org/docs/cli/#update 
the update has added some new packages as well, blinker and typing-extensions. All still works and is passing so perhaps its fine

MebinAbraham: I think that would still update it. Are you referring to "Note that this will not update versions for dependencies outside their version constraints specified in the pyproject.toml file."

Even with this, it is still going to update as long as the constraints are met.

katie-gardner: Exactly that yep, but then as you say it'd have to be specific version numbers

katie-gardner (approved):
As all tests are still passing, don't see an issue with other updates

PR #200, opened 2023-06-01
Add rules for validation of calculated sum of dynamic answers

petechd:
New validation rules for calculated validation of dynamic answers.

### How to review
Test against new schema in runner. Refer to the card.

### Checklist


petechd:
[Looping 3.5 - Validating dynamic answers based on a list collector (3 SP)](https://trello.com/c/WPWjw3Vp/5324-looping-35-validating-dynamic-answers-based-on-a-list-collector-3-sp)

Yuyuutsu (approved):
LGTM ! and runs as expected on the validating dynamic runner branch

berroar (commented):
Is this being tested in validator? do we need to add valid/invalid schema to test this?

petechd:
> Is this being tested in validator? do we need to add valid/invalid schema to test this?

Two test schemas added, valid and invalid, also some extra validation for when answers to calculate is only one item.

tests/schemas/invalid/test_invalid_validation_sum_against_total_dynamic_answers.json
```diff
@@ -0,0 +1,253 @@
...
+                    "calculation_type": "sum",
+                    "value": 1000,
+                    "answers_to_calculate": [
+                      "percentage-of-shopping-elsewhere"
+                    ],
+                    "conditions": ["equals"]
+                  }
+                ],
+                "id": "dynamic-answer-only-question",
+                "title": "How much do you spend at each of the following supermarket?",
```
katie-gardner, line 244: fails as expected if I put this in the valid schemas folder but I think you need a test that loads this schema and asserts that you get the correct Answers too short message when on this block
at the moment when I run unit tests the coverage report doesn't include the new type of error

katie-gardner (changes requested):
Looks good! just missing a test for the invalid schema 👍

tests/schemas/invalid/test_invalid_validation_sum_against_total_dynamic_answers.json
petechd: Added a test now, thanks.

schemas/questions/types/calculated.json
```diff
@@ -59,7 +89,7 @@
             },
             "answers_to_calculate": {
               "type": "array",
-              "minItems": 2,
+              "minItems": 1,
```
kylelawsonAND, line 92: Based on the changes to `calculated_question_validator.py`, will this now allow a list of non-dynamic sources with a length of 1?

katie-gardner (changes requested):
Suggested a test simplification but otherwise all good!

tests/validators/blocks/test_calculated_question_validator.py
```diff
@@ -68,3 +68,57 @@ def test_invalid_id_in_answers_to_calculate():
...
 
     assert expected_error_messages == validator.errors
+
+
+def test_answers_to_calculate_too_short():
+    filename = (
+        "schemas/invalid/test_invalid_validation_sum_against_total_dynamic_answers.json"
+    )
+    schema = QuestionnaireSchema(_open_and_load_schema_file(filename))
+    question = {
```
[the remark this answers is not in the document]
petechd: Added!

katie-gardner (approved):
Looking good! 👍

schemas/questions/types/calculated.json
petechd: It will fail validation in python if it's static answers ids list with len of 1 now. So the schema won't get through and fail.

rmccar (approved):

kylelawsonAND (approved):

PR #201, opened 2023-06-06
Looping 4 - Support calculated summaries having a single dynamic answer

katie-gardner:
This PR updates validation to allow a calculated summary to have just one item in the list to add up **if** that item references a dynamic-answer, that way it still resolves to a list

Dynamic and static answers can be used together on a calculated summary, which did not require any changes.

([Looping 4 feature branch](https://github.com/ONSdigital/eq-questionnaire-runner/compare/feature-prepop...feat-looping-4))

### How to review
Read through the requirements in the trello card and ensure the changes are correct and make sense. Check that you can add a calculated summary to a schema with a single item which is a dynamic-answer, and that you can't add a single item that is a static-answer.

Make sure tests fully cover the changes

### Checklist

### DIFF
--- a/tests/schemas/invalid/test_invalid_new_calculated_summary_with_dynamic_answers.json
+++ b/tests/schemas/invalid/test_invalid_new_calculated_summary_with_dynamic_answers.json
@@ -47,57 +47,6 @@
           "id": "group",
           "blocks": [
             {
-              "type": "ListCollectorDrivingQuestion",
-              "id": "any-transport",
-              "for_list": "transport",
-              "question": {
-                "type": "General",
-                "id": "any-transport-question",
-                "title": "Do you use public transport?",
-                "answers": [
-                  {
-                    "type": "Radio",
-                    "id": "any-transport-answer",
-                    "mandatory": true,
-                    "options": [
-                      {
-                        "label": "Yes",
-                        "value": "Yes",
-                        "action": {
-                          "type": "RedirectToListAddBlock",
-                          "params": {
-                            "block_id": "add-transport",
-                            "list_name": "transport"
-                          }
-                        }
-                      },
-                      {
-                        "label": "No",
-                        "value": "No"
-                      }
-                    ]
-                  }
-                ]
-              },
-              "routing_rules": [
-                {
-                  "section": "End",
-                  "when": {
-                    "==": [
-                      {
-                        "source": "answers",
-                        "identifier": "any-transport-answer"
-                      },
-                      "No"
-                    ]
-                  }
-                },
-                {
-                  "block": "list-collector"
-                }
-              ]
-            },
-            {
               "id": "list-collector",
               "type": "ListCollector",
               "for_list": "transport",
--- a/tests/schemas/valid/test_new_calculated_summary_with_dynamic_answers.json
+++ b/tests/schemas/valid/test_new_calculated_summary_with_dynamic_answers.json
@@ -47,57 +47,6 @@
           "id": "group",
           "blocks": [
             {
-              "type": "ListCollectorDrivingQuestion",
-              "id": "any-transport",
-              "for_list": "transport",
-              "question": {
-                "type": "General",
-                "id": "any-transport-question",
-                "title": "Do you use public transport?",
-                "answers": [
-                  {
-                    "type": "Radio",
-                    "id": "any-transport-answer",
-                    "mandatory": true,
-                    "options": [
-                      {
-                        "label": "Yes",
-                        "value": "Yes",
-                        "action": {
-                          "type": "RedirectToListAddBlock",
-                          "params": {
-                            "block_id": "add-transport",
-                            "list_name": "transport"
-                          }
-                        }
-                      },
-                      {
-                        "label": "No",
-                        "value": "No"
-                      }
-                    ]
-                  }
-                ]
-              },
-              "routing_rules": [
-                {
-                  "section": "End",
-                  "when": {
-                    "==": [
-                      {
-                        "source": "answers",
-                        "identifier": "any-transport-answer"
-                      },
-                      "No"
-                    ]
-                  }
-                },
-                {
-                  "block": "list-collector"
-                }
-              ]
-            },
-            {
               "id": "list-collector",
               "type": "ListCollector",
               "for_list": "transport",


katie-gardner:
[Looping 4 - Calculated summary using a list of repeating answers (5 SP)](https://trello.com/c/jXpOixe8/5320-looping-4-calculated-summary-using-a-list-of-repeating-answers-5-sp)

kylelawsonAND (approved):
👍

app/validators/questionnaire_schema.py
```diff
@@ -360,6 +360,14 @@ def get_all_answer_ids(self, block_id):
             for answer in self.get_answers_from_question(question)
         }
 
+    def get_all_dynamic_answer_ids(self, block_id):
```
petechd, line 363: We probably need to sync our PRs and use the same method in both so there is no duplication of methods, I will use your method as it has better name and it's more universal.

katie-gardner: Ahh good point - thanks Pete 👍

app/validators/blocks/calculated_summary_block_validator.py
```diff
@@ -19,13 +20,34 @@ def validate(self):
...
+        single_answer_id = answers[0]["id"]
+        question_block_id = self.questionnaire_schema.get_block_id_by_answer_id(
+            single_answer_id
+        )
+        if (
+            single_answer_id
+            not in self.questionnaire_schema.get_all_dynamic_answer_ids(
+                question_block_id
+            )
+        ):
```
petechd, line 45:
```suggestion
        if len(answers) == 1 and answers[
                0
            ] not in self.questionnaire_schema.get_all_dynamic_answer_ids(
                self.block["id"]
            ):
```

I would make it more compacted, especially when you have self.block unlike me in my PR, docstring explains it already. I don't know what others think?

schemas/rules/operators/numeric/sum.json
```diff
@@ -8,7 +8,7 @@
     "properties": {
       "+": {
         "type": "array",
-        "minItems": 2,
+        "minItems": 1,
```
petechd, line 11: This probably means grand calculated summary allows one item now (as this operator rule is used by both GCS and CS)? Do we need to add extra python validation for GCS?

tests/schemas/invalid/test_invalid_new_calculated_summary_with_dynamic_answers.json
```diff
@@ -0,0 +1,332 @@
+{
+  "mime_type": "application/json/ons/eq",
+  "language": "en",
+  "schema_version": "0.0.1",
+  "data_version": "0.0.3",
+  "survey_id": "0",
+  "title": "Calculated Summary with Dynamic Answers",
+  "theme": "default",
+  "description": "demo of invalid calculated summary which references a single static answer from a question with both static and dynamic answers.",
```
petechd, line 9:
```suggestion
  "description": "Demo of invalid calculated summary which references a single static answer from a question with both static and dynamic answers.",
```

app/validators/blocks/calculated_summary_block_validator.py
```diff
@@ -6,6 +6,7 @@ class CalculatedSummaryBlockValidator(CalculationBlockValidator):
         "Answer ids for calculated summary must be set before calculated summary block"
     )
     ANSWER_SET_IN_DIFFERENT_SECTION_FOR_CALCULATED_SUMMARY = "Answer ids for calculated summary must be set in the same section as the calculated summary block"
+    CALCULATED_SUMMARY_WITH_NON_DYNAMIC_SINGLE_ANSWER = "Calculated summaries cannot consist of a single answer unless it is a dynamic answer"
```
petechd, line 9: We can add {answer_id} here so it's easier to track down and fix when validation fails.

app/validators/blocks/calculated_summary_block_validator.py
katie-gardner: Yeah good point - can definitely make it shorter, though I just realised that using `self.block["id"]` isn't working as that is the block id of the calculated summary rather than the block id of the question for the single answer, I think its going to be the same problem on your PR

schemas/rules/operators/numeric/sum.json
katie-gardner: Luckily its ok for GCS, if you look at `sum.json` the gcs values come from `sum_calculated_summary_values` which still has a minimum of 2

app/validators/blocks/calculated_summary_block_validator.py
katie-gardner: 👍 I added answer_id to the error message in the same way as the block one, so it'd be similar to other CS messages, if that seems OK?

app/validators/blocks/calculated_summary_block_validator.py
katie-gardner: I kept `single_answer_id` as its used twice and I think it helps with readability, but made the rest more compact as per suggestion!

tests/schemas/valid/test_new_calculated_summary_with_dynamic_answers.json
```diff
@@ -0,0 +1,322 @@
+{
```
petechd, line 1: We could remove the blocks we don't care about for the tests, to reduce the size of this schema, are there any blocks here that are irrelevant? Same with the other schema.

berroar (approved):

tests/schemas/valid/test_new_calculated_summary_with_dynamic_answers.json
katie-gardner: I'm not sure which blocks I can remove without making it an unrealistic schema, I need the add/edit/remove and dynamic answers to have the right setup for the calculated summary, are there any you spotted in particular that I don't need?

petechd (approved):

katie-gardner:
[Looping 4.5 - Calculated summary using dynamic answers (3 SP)](https://trello.com/c/8njslssz/5721-looping-45-calculated-summary-using-dynamic-answers-3-sp)

tests/schemas/valid/test_new_calculated_summary_with_dynamic_answers.json
petechd: @katie-gardner I probably saw that list collector driving questions in both schemas, they are not needed, since we only validate unrelated changes.

katie-gardner: @petechd - sure, sounds good, removed

PR #203, opened 2023-06-08
Adding support for repeating blocks in list collector

kylelawsonAND:
This PR supports the implemention of the feature [ Looping 2.5](https://trello.com/c/FYXB2EYJ), the runner PR of which is [here](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1106).
To support list collector having repeating blocks, there must be a new valid key "repeating_blocks at the same level as "add_block", "edit_block" and "remove_block", which is an array of blocks with at least one entry. These blocks must support placeholders and all of the usual question types, but skipping and routing is not yet supported. 

### Changes
**app/validators/questionnaire_schema.py**: Public method `get_context_from_match` will get the repeating block from the match path via `get_element_value` for the question that the given match is a reference to, if it belongs to a repeating block, else it finds the parent block as normal.
`QuestionnaireSchema` has a new class attribute `repeating_blocks` which records the repeating blocks in the schema much like blocks and sub_blocks. 
`blocks_by_id` now records the `repeating_blocks` ids as well as `blocks` and `sub_blocks`. 

**app/validators/sections/section_validator.py**: Public method `validate_blocks` will validate repeating blocks if they exist as children of the current block. It runs standard block validation as well as the owned methods `validate_question` and `validate_variants`.

**schemas/blocks/list_collector.json**: Added a new non-mandatory key "repeating_blocks" to the list collector schema. The value is the new `schemas/blocks/list_repeating_blocks.json` schema.

**schemas/blocks/list_repeating_blocks.json**: This is a new schema to define the array of repeating blocks, where there must be at leats one entry of the new `schemas/blocks/list_repeating_block.json` schema. 

**schemas/blocks/list_repeating_block.json**: A new block schema for the repeating blocks. This block must have id, type and question, and optionally page_title. The type must also be `ListRepeatingBlock`.

**tests/schemas/invalid/test_invalid_list_collector_repeating_blocks_empty.json**: An invalid schema which violates the new ajv schema validation rule where the repeating)blocks array must have one entry. 

**tests/schemas/invalid/test_invalid_list_collector_repeating_blocks_placeholder_references_same_block.json**: An invalid schema which violates validator's placeholder rules such that a repeating block placeholder references a value from the same block. 

**tests/schemas/valid/test_list_collector_repeating_blocks.json**: A valid list collector with repeating blocks schema. 

**tests/validators/blocks/test_block_validator.py**: A test to ensure the changes to validator to support validating the repeating blocks is working, in particular the placeholder same block reference error is produced. 

### How to review
Familiarise yourself with the Looping 2.5 card and scenarios linked above. Should anything else be added to validator to support the repeating blocks feature?
All new and existing unit, integration and functional tests should be passing locally as well as on PR.

### DIFF
--- a/app/validators/blocks/calculated_summary_block_validator.py
+++ b/app/validators/blocks/calculated_summary_block_validator.py
@@ -6,6 +6,7 @@ class CalculatedSummaryBlockValidator(CalculationBlockValidator):
         "Answer ids for calculated summary must be set before calculated summary block"
     )
     ANSWER_SET_IN_DIFFERENT_SECTION_FOR_CALCULATED_SUMMARY = "Answer ids for calculated summary must be set in the same section as the calculated summary block"
+    CALCULATED_SUMMARY_WITH_NON_DYNAMIC_SINGLE_ANSWER = "Calculated summaries cannot consist of a single answer unless it is a dynamic answer"
 
     def __init__(self, block, questionnaire_schema):
         super().__init__(block, questionnaire_schema)
@@ -19,12 +20,31 @@ class CalculatedSummaryBlockValidator(CalculationBlockValidator):
         if (answers := self.get_answers(self.answers_to_calculate)) is None:
             return self.errors
 
+        self.validate_single_answer_is_for_dynamic_answers(answers)
         self.validate_answer_id_set_before_calculated_summary_block()
         self.validate_answer_id_for_calculated_summary_not_in_different_section()
 
         self.validate_answer_types(answers)
 
         return self.errors
+
+    def validate_single_answer_is_for_dynamic_answers(self, answers: list[dict]):
+        """Validate that if there is only one answer in the answers_to_calculate list, it's for dynamic answers"""
+        if len(answers) == 1:
+            single_answer_id = answers[0]["id"]
+            if (
+                single_answer_id
+                not in self.questionnaire_schema.get_all_dynamic_answer_ids(
+                    self.questionnaire_schema.get_block_id_by_answer_id(
+                        single_answer_id
+                    )
+                )
+            ):
+                self.add_error(
+                    self.CALCULATED_SUMMARY_WITH_NON_DYNAMIC_SINGLE_ANSWER,
+                    block_id=self.block["id"],
+                    answer_id=single_answer_id,
+                )
 
     def validate_answer_id_set_before_calculated_summary_block(self):
         for answer_id in self.answers_to_calculate:
--- a/app/validators/placeholders/placeholder_validator.py
+++ b/app/validators/placeholders/placeholder_validator.py
@@ -141,26 +141,35 @@ class PlaceholderValidator(Validator):
             return None
 
         value = arguments["value"]
-        answer_id = value.get("identifier")
+        identifier = value.get("identifier")
         unit = arguments["unit"]
+        source_answer_ids = []
 
-        if self.errors or (
-            unit
-            == self.questionnaire_schema.answers_with_context[answer_id]["answer"][
-                "unit"
-            ]
-        ):
-            return None
+        if value.get("source") == "calculated_summary":
+            if calculated_summary_source_answer_ids := self.questionnaire_schema.get_calculation_block_ids(
+                block=self.questionnaire_schema.get_block(identifier),
+                source_type="answers",
+            ):
+                source_answer_ids = calculated_summary_source_answer_ids
 
-        self.add_error(
-            error_messages.ANSWER_UNIT_AND_TRANSFORM_UNIT_MISMATCH.format(
-                answer_unit=self.questionnaire_schema.answers_with_context[answer_id][
-                    "answer"
-                ]["unit"],
-                transform_unit=unit,
-            ),
-            identifier=answer_id,
-        )
+        else:
+            source_answer_ids.append(identifier)
+
+        for source_answer_id in source_answer_ids:
+
+            answer_unit = self.questionnaire_schema.answers_with_context[
+                source_answer_id
+            ]["answer"]["unit"]
+
+            if unit != answer_unit:
+                self.add_error(
+                    error_messages.ANSWER_UNIT_AND_TRANSFORM_UNIT_MISMATCH.format(
+                        answer_unit=answer_unit,
+                        transform_unit=unit,
+                    ),
+                    identifier=identifier,
+                )
+                break
 
     def _validate_placeholder_previous_transforms(self, transforms):
         # First transform can't reference a previous transform
--- a/app/validators/questionnaire_schema.py
+++ b/app/validators/questionnaire_schema.py
@@ -366,6 +366,7 @@ class QuestionnaireSchema:
             for answer in self.get_answers_from_question(question)
         }
 
+    @lru_cache
     def get_all_dynamic_answer_ids(self, block_id):
         questions = self.get_all_questions_for_block(self.blocks_by_id[block_id])
         return {
--- a/schemas/rules/operators/numeric/sum.json
+++ b/schemas/rules/operators/numeric/sum.json
@@ -8,7 +8,7 @@
     "properties": {
       "+": {
         "type": "array",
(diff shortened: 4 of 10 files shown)


kylelawsonAND:
[Capture of related answers on separate pages - Looping 2.5 (5 SP)](https://trello.com/c/FYXB2EYJ/5502-capture-of-related-answers-on-separate-pages-looping-25-5-sp)

app/validators/sections/section_validator.py
```diff
@@ -96,6 +96,15 @@ def validate_blocks(self, group_id):
...
             self.validate_variants(block)
+            # Repeating blocks must be validated here instead of from ListCollectorValidator as the latter cannot do standard block validation
+            for repeating_block in block.get("repeating_blocks", []):
+                block_validator = get_block_validator(
+                    repeating_block, self.questionnaire_schema
+                )
+                self.errors += block_validator.validate()
+
+                self.validate_question(repeating_block)
+                self.validate_variants(repeating_block)
```
petechd, line 107: I think it would be cleaner to read and follow if you do `self.validate_repeating_blocks(block)`.

tests/validators/blocks/test_block_validator.py
```diff
@@ -93,3 +93,22 @@ def test_invalid_self_reference():
         }
     ]
     assert validator.errors == expected_errors
+
+
+def test_invalid_list_collector_repeating_blocks_placeholder_references_same_block():
```
petechd, line 98: I think we need to add a test of `SectionValidator` since that's where you added your new validation.

app/validators/sections/section_validator.py
```diff
@@ -96,6 +96,15 @@ def validate_blocks(self, group_id):
 
             self.validate_question(block)
             self.validate_variants(block)
+            # Repeating blocks must be validated here instead of from ListCollectorValidator as the latter cannot do standard block validation
```
petechd, line 99: Can you not do this code in list collector validator? something like:

```
for repeating_block in self.block.get("repeating_blocks", []):
    section_validator = SectionValidator(repeating_block, self.questionnaire_schema)
    block_validator = get_block_validator(
        repeating_block, self.questionnaire_schema
    )
    self.errors += block_validator.validate()

    section_validator.validate_question(repeating_block)
    section_validator.validate_variants(repeating_block)
```

kylelawsonAND: This causes a circular dependancy - it's also passing a block to the section parameter so feel like a misuse. 
I think moving RB validation to ListCollectorValidator could be done by moving lots of SectionValidator out in a generic validator helper, but feels out of scope here for the scale of changes it'd be.

tests/validators/blocks/test_block_validator.py
kylelawsonAND: Moved the test from test_block_validator to test_section_validator and calling validation from an instance of SectionValidator 👍

app/validators/sections/section_validator.py
kylelawsonAND: Made a function which is explicitly only called for list collector block validators

schemas/blocks/list_collector.json
```diff
@@ -45,6 +45,9 @@
       "add_block": {
         "$ref": "list_add_edit_remove_question.json#/block"
       },
+      "repeating_blocks": {
```
katie-gardner, line 48: Looking at `answer_summary.json` which has the answer_ids as a list inline, I'm vaguely thinking it would be simpler and clearer to just have that this is an array here with minItems one, referencing the repeating block, and not have a separate `list_repeating_blocks.json` file

katie-gardner (changes requested):
Two minor comments, but nothing else from me, this looks really good 👍

app/validators/sections/section_validator.py
petechd: I think it is better if we do this on the method level, so it is same  everywhere, same as we do in `self.validate_question()`:
So:
```python
question = block_or_variant.get("question")
        if question:
```
Something like:
```python
repeating_blocks = block.get("repeating_blocks", []):
     if repeating_blocks:
```
Thoughts?

kylelawsonAND: Yep agree that would be clean and consistent, have updated to reflect this. I have also removed the check for the block validator being a ListCollectorValidator as it seemed redundant given the following method is checking for the repeating blocks and only validating if they exist - only list collectors would have them.

tests/schemas/invalid/test_invalid_list_collector_repeating_blocks_empty.json
```diff
@@ -0,0 +1,180 @@
+{
+  "mime_type": "application/json/ons/eq",
+  "language": "en",
+  "schema_version": "0.0.1",
```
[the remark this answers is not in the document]
kylelawsonAND: It's there for testing against AJV validation, which it would fail as expected. But I don't believe we automate AJV validation unless I missed that.

app/validators/sections/section_validator.py
katie-gardner: I like this change 👍

tests/schemas/invalid/test_invalid_list_collector_repeating_blocks_empty.json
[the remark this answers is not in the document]
katie-gardner: Hmmm I'm not sure if this is worth having if it only fails AJV when manually tested, but probably something for the team to decide

schemas/blocks/list_collector.json
kylelawsonAND: Great shout, done 👍

katie-gardner (approved):
Points all addressed and looking good 👍

berroar (approved):

tests/validators/sections/test_section_validator.py
```diff
@@ -1,4 +1,5 @@
 from app import error_messages
```
petechd, line 1: From the card: "We will only ever have one list collector per list", we need to add such validation.

schemas/blocks/list_repeating_block.json
```diff
@@ -0,0 +1,24 @@
...
+        "$ref": "../common_definitions.json#/identifier"
+      },
+      "page_title": {
+        "$ref": "../common_definitions.json#/non_empty_string"
+      },
+      "type": {
+        "type": "string",
+        "enum": ["ListRepeatingQuestion"]
+      },
+      "question": {
```
petechd, line 17: Are we supposed to support all kinds of questions (types)?

petechd: Or do we need a separate definition for repeating block question?

tests/validators/sections/test_section_validator.py
kylelawsonAND: Ah fair point, working on this now

schemas/blocks/list_repeating_block.json
kylelawsonAND: As far as I'm aware it's any question, as the limitations on the question weren't specified, and Mebin was fine with this impl of any question

tests/validators/sections/test_section_validator.py
kylelawsonAND: Validation for this now added, covers multiple list collectors referring to lists with repeating blocks, within the same section, and in different sections.

schemas/blocks/list_repeating_block.json
petechd: Can you make sure and ask Donna if it's all types of questions?

kylelawsonAND: Sure, this will be blocked until Monday though as she's on leave

kylelawsonAND: Donna cleared this - these question types are allowed 👍

Yuyuutsu (approved):

PR #204, opened 2023-06-16
Fix validation of format_unit transform using calculated summary source

petechd:

### How to review

### Checklist

### DIFF
--- a/app/validators/placeholders/placeholder_validator.py
+++ b/app/validators/placeholders/placeholder_validator.py
@@ -144,19 +144,23 @@ class PlaceholderValidator(Validator):
         answer_id = value.get("identifier")
         unit = arguments["unit"]
 
-        if self.errors or (
-            unit
-            == self.questionnaire_schema.answers_with_context[answer_id]["answer"][
-                "unit"
-            ]
-        ):
+        if value.get("source") == "calculated_summary":
+            answers_unit = (
+                self.questionnaire_schema.get_calculated_summary_source_answer_ids(
+                    answer_id
+                )
+            )
+        else:
+            answers_unit = self.questionnaire_schema.answers_with_context[answer_id][
+                "answer"
+            ]["unit"]
+
+        if self.errors or (unit == answers_unit):
             return None
 
         self.add_error(
             error_messages.ANSWER_UNIT_AND_TRANSFORM_UNIT_MISMATCH.format(
-                answer_unit=self.questionnaire_schema.answers_with_context[answer_id][
-                    "answer"
-                ]["unit"],
+                answer_unit=answers_unit,
                 transform_unit=unit,
             ),
             identifier=answer_id,
--- a/app/validators/questionnaire_schema.py
+++ b/app/validators/questionnaire_schema.py
@@ -517,3 +517,11 @@ class QuestionnaireSchema:
             *question.get("dynamic_answers", {}).get("answers", []),
             *question.get("answers", []),
         ]
+
+    def get_calculated_summary_source_answer_ids(self, block_id):
+        operations = self.get_block(block_id)["calculation"]["operation"]
+        for operation in operations:
+            for item in operations[operation]:
+                if item["source"] == "answers":
+                    answer = self.get_answer(item["identifier"])
+                    return answer["unit"]
--- a/tests/test_placeholder_validator.py
+++ b/tests/test_placeholder_validator.py
@@ -189,3 +189,24 @@ def test_validation_answer_and_transform_unit_match(
         arguments=arguments, transform_type=transform_type
     )
     assert validator.errors == expected_error
+
+
+@pytest.mark.parametrize(
+    "expected_error",
+    [
+        [
+            {
+                "message": error_messages.ANSWER_UNIT_AND_TRANSFORM_UNIT_MISMATCH.format(
+                    answer_unit="length-centimeter", transform_unit="length-meter"
+                ),
+                "identifier": "unit-total-playback",
+            }
+        ]
+    ],
+)
+def test_validation_answer_and_transform_unit_mismatch(expected_error):
+    filename = "schemas/invalid/test_invalid_calculated_summary_units_mismatch.json"
+    schema_file = _open_and_load_schema_file(filename)
+    validator = PlaceholderValidator(schema_file)
+    validator.validate()
+    assert validator.errors == expected_error
--- a/tests/schemas/invalid/test_invalid_calculated_summary_units_mismatch.json
+++ b/tests/schemas/invalid/test_invalid_calculated_summary_units_mismatch.json
@@ -1,0 +1,127 @@
+{
+  "mime_type": "application/json/ons/eq",
+  "language": "en",
+  "schema_version": "0.0.1",
+  "data_version": "0.0.3",
+  "survey_id": "0",
+  "title": "A test schema to demo Calculated Summary",
+  "theme": "default",
+  "description": "A schema to demo Calculated Summary usage in value source.",
+  "metadata": [
+    {
+      "name": "user_id",
+      "type": "string"
+    },
+    {
+      "name": "period_id",
+      "type": "string"
+    },
+    {
+      "name": "ru_name",
+      "type": "string"
+    }
+  ],
+  "questionnaire_flow": {
+    "type": "Linear",
+    "options": {
+      "summary": {
+        "collapsible": false
+      }
+    }
+  },
+  "sections": [
+    {
+      "id": "default-section",
+      "groups": [
+        {
+          "id": "group",
+          "title": "Total a range of values",
+          "blocks": [
+            {
+              "type": "Question",
+              "id": "first-number-block",
+              "question": {
+                "id": "first-number-question",
+                "title": "First Number Question Title",
+                "type": "General",
+                "answers": [
+                  {
+                    "id": "first-number-answer-unit-total",
+                    "label": "Second answer label in unit total",
+                    "mandatory": true,
+                    "type": "Unit",
+                    "unit_length": "short",
+                    "unit": "length-centimeter"
+                  },
+                  {
+                    "id": "second-number-answer-unit-total",
+                    "label": "Third answer label in unit total",
+                    "mandatory": true,
+                    "type": "Unit",
+                    "unit_length": "short",
+                    "unit": "length-centimeter"
+                  }
+                ]
+              }
+            },
+            {
+              "type": "CalculatedSummary",
+              "id": "unit-total-playback",
+              "title": "We calculate the total of unit values entered to be %(total)s. Is this correct?",
+              "page_title": "Total Unit Values",
+              "calculation": {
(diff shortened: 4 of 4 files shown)


petechd:
[Bug: Validator errors for placeholders using units with calculated summary sources](https://trello.com/c/j8To6SiI/5691-bug-validator-errors-for-placeholders-using-units-with-calculated-summary-sources)

app/validators/questionnaire_schema.py
```diff
@@ -517,3 +517,11 @@ def get_answers_from_question(question):
             *question.get("dynamic_answers", {}).get("answers", []),
             *question.get("answers", []),
         ]
+
+    def get_calculated_summary_source_answer_ids(self, block_id):
```
katie-gardner, line 521: the function is called `get_calculated_summary_source_answer_ids` but it returns the "unit" of the first answer, so I think it should be renamed to `get_calculated_summary_unit_type`

I think you might not need this though - see other comment where you could reuse `self.get_calculation_block_ids`

tests/test_placeholder_validator.py
```diff
@@ -189,3 +189,24 @@ def test_validation_answer_and_transform_unit_match(
         arguments=arguments, transform_type=transform_type
     )
     assert validator.errors == expected_error
+
+
+@pytest.mark.parametrize(
```
katie-gardner, line 194: since there is only one test case - I don't think theres any need to parametrize this, the expected errors can just be in the function

app/validators/placeholders/placeholder_validator.py
```diff
@@ -144,19 +144,23 @@ def validate_answer_and_transform_unit_match(self, *, arguments, transform_type)
         answer_id = value.get("identifier")
```
katie-gardner, line 144: in the case of the value source being a calculated summary, `value.get("identifier")` is the block_id for the calculated summary, so I think the check should be here and existing code calling it `answer_id` should change

if you reuse the `get_calculation_block_ids` from questionnaire schema you could do something like this and then the new method wouldn't be needed
```
if value.get("source") == "calculated_summary":
    if answer_ids := self.questionnaire_schema.get_calculation_block_ids(
        block=self.questionnaire_schema.get_block(value.get("identifier")),
        source_type="answers",
    ):
        answer_id = answer_ids[0]
```

then existing answers_unit = ... would work

katie-gardner (changes requested):

tests/test_placeholder_validator.py
petechd: I thinking here about the future and doing it with parameters if we want to expand, similar to the one above but ok will change it now.

app/validators/placeholders/placeholder_validator.py
```diff
@@ -144,19 +144,26 @@ def validate_answer_and_transform_unit_match(self, *, arguments, transform_type)
...
+            core_answer_id = value.get("identifier")
+
+        answers_unit = self.questionnaire_schema.answers_with_context[core_answer_id][
+            "answer"
+        ]["unit"]
+
+        if self.errors or (unit == answers_unit):
             return None
 
         self.add_error(
```
kylelawsonAND, line 164: Uploading Screen Recording 2023-06-19 at 11.36.20.mov…

kylelawsonAND: https://github.com/ONSdigital/eq-questionnaire-validator/assets/108286538/27c36514-2d68-43cb-a77c-726d2d297afb

kylelawsonAND: second time's the charm

petechd: Ok, seems important to fix this so I've changed the logic and added extra tests so we know it works for all scenarios above now. Thanks!

app/validators/placeholders/placeholder_validator.py
katie-gardner: agree with passing it into the error at the end 👍 maybe it could be renamed from `answer_id` to `identifier` though? in your new test the `answer_id` is `unit-total-playback` which is the block id for the calculated summary

app/validators/placeholders/placeholder_validator.py
```diff
@@ -143,24 +143,32 @@ def validate_answer_and_transform_unit_match(self, *, arguments, transform_type)
...
-        if self.errors or (
-            unit
-            == self.questionnaire_schema.answers_with_context[answer_id]["answer"][
-                "unit"
-            ]
-        ):
-            return None
+        if value.get("source") == "calculated_summary":
+            if calculated_summary_source_answer_ids := self.questionnaire_schema.get_calculation_block_ids(
+                block=self.questionnaire_schema.get_block(value.get("identifier")),
```
katie-gardner, line 150: if you go with `identifier = value.get("identifier")` above, you can reuse that here

app/validators/placeholders/placeholder_validator.py
```diff
@@ -143,24 +143,32 @@ def validate_answer_and_transform_unit_match(self, *, arguments, transform_type)
...
-            error_messages.ANSWER_UNIT_AND_TRANSFORM_UNIT_MISMATCH.format(
-                answer_unit=self.questionnaire_schema.answers_with_context[answer_id][
-                    "answer"
-                ]["unit"],
-                transform_unit=unit,
-            ),
-            identifier=answer_id,
-        )
+        else:
+            source_answer_ids.append(value.get("identifier"))
```
katie-gardner, line 156: can reuse here too

tests/test_placeholder_validator.py
```diff
@@ -189,3 +190,62 @@ def test_validation_answer_and_transform_unit_match(
...
+    filename = "schemas/invalid/test_invalid_calculated_summary_units_mismatch.json"
+    schema_file = _open_and_load_schema_file(filename)
+    validator = PlaceholderValidator(schema_file)
+    validator.validate()
+    assert validator.errors == [
+        {
+            "message": error_messages.ANSWER_UNIT_AND_TRANSFORM_UNIT_MISMATCH.format(
+                answer_unit="length-centimeter", transform_unit="length-meter"
+            ),
+            "identifier": "unit-total-playback",
```
katie-gardner, line 205: as the two error messages are identical, would another parameter for the mismatching answer id be a good idea here? so we'd have identifier and answer_id

tests/test_placeholder_validator.py
```diff
@@ -189,3 +190,62 @@ def test_validation_answer_and_transform_unit_match(
...
+        },
+    ]
+
+
+def test_validation_answer_and_transform_unit_single_source_answer_mismatch():
+    filename = "schemas/invalid/test_invalid_calculated_summary_units_mismatch.json"
+    schema_file = _open_and_load_schema_file(filename)
+    schema = QuestionnaireSchema(schema_file)
+    answer = schema.get_answer("second-number-answer-unit-total")
```
katie-gardner, line 220: you could probably parametrize these two tests with an `answer_id` parameter as its the only thing that changes

app/validators/placeholders/placeholder_validator.py
```diff
@@ -143,24 +143,32 @@ def validate_answer_and_transform_unit_match(self, *, arguments, transform_type)
...
+                source_answer_id
+            ]["answer"]["unit"]
+
+            if unit != answers_unit:
+                self.add_error(
+                    error_messages.ANSWER_UNIT_AND_TRANSFORM_UNIT_MISMATCH.format(
+                        answer_unit=answers_unit,
+                        transform_unit=unit,
+                    ),
+                    identifier=answer_id,
```
katie-gardner, line 170: as mentioned below, if `answer_id` was changed to identifier, then you could separately have `answer_id=source_answer_id` included in the error

katie-gardner (changes requested):

tests/test_placeholder_validator.py
katie-gardner: sorry, now that you've added more tests it does make sense 😁

tests/test_placeholder_validator.py
petechd: It is slightly more complicated, I had to add answer_id and mismatched anwer_id as parameters, since we add the change from this comment: https://github.com/ONSdigital/eq-questionnaire-validator/pull/204#discussion_r1234057634

petechd: We might have to keep it as it was, cause if the source is answer then error message has duplicates:
`{'message': "The answer unit and transform unit mismatch, 'mile' not 'meter'", 'identifier': 'average-distance', 'answer_id': 'average-distance'}`

petechd: Check https://github.com/ONSdigital/eq-questionnaire-validator/pull/204#discussion_r1234859166

petechd: We are already in a loop for a single answer id (when the error happens), so maybe we should keep it as it is to not overcomplicate it and keep it similar to other errors? We would have to build a list to pass into the final message, wouldn't this be over-engineered? I think it's more like errors in validator where you add one for each occurrence but I am happy to change it.

petechd: Also for calculated it would have to be a different error message instead of answer_id field it would be answer_ids.

petechd: The question is: when you have a mismatch, do you want an error message for each mismatch?

katie-gardner: hmm, I see the point, I just don't like that we've got two absolutely identical error messages
```
[{'identifier': 'unit-total-playback',
  'message': "The answer unit and transform unit mismatch, 'length-centimeter' "
             "not 'length-meter'"},
 {'identifier': 'unit-total-playback',
  'message': "The answer unit and transform unit mismatch, 'length-centimeter' "
             "not 'length-meter'"}]
```
because each answer caused another error, but they're in the same block and only that is included in the message

If you omit `answer_id` can the error show just once? perhaps intead of iterating over answers and adding the error, you could do a list comprehension that gets the units, and then if any unit doesn't match the transform one, you just add a single error

katie-gardner: I think an error for each mismatch is helpful if we know why each mismatch is different or what each one is, otherwise for duplicate errors I think its better to only show it once

petechd:
Removed duplication of error messages when calculated summary. Identifier field is now block id if calculated summary is the source of the format or answer id, if answer value source.

app/validators/placeholders/placeholder_validator.py
kylelawsonAND: Nice one 👌

katie-gardner (approved):
Working locally and no more duplicate errors 👍

tests/test_placeholder_validator.py
petechd: Discussed in ad-hoc session.

kylelawsonAND (approved):
All looks good to me now 🎉

berroar (approved):

PR #205, opened 2023-07-05
Add list collector content validation

petechd:
Introduces new `List collector content` type of block to definitions, added all the required fields. New schema in runner added for validation.

### How to review
Validate correct schema in runner branch relevant to this change.

### Checklist


petechd:
[Support for non question list collector (3 SP)](https://trello.com/c/acXLlgAj/5704-support-for-non-question-list-collector-3-sp)

schemas/blocks/list_collector_content.json
```diff
@@ -0,0 +1,64 @@
...
+    "type": "object",
+    "properties": {
+      "id": {
+        "$ref": "../common_definitions.json#/identifier"
+      },
+      "type": {
+        "type": "string",
+        "enum": ["ListCollectorContent"]
+      },
+      "content": {
```
katie-gardner, line 27: The ticket suggests this should support both contents and content variants

schemas/blocks/list_collector_content.json
```diff
@@ -0,0 +1,64 @@
...
+        "items": {
+          "$ref": "list_repeating_block.json#/block"
+        }
+      },
+      "summary": {
+        "$ref": "list_collector_content.json#/summary"
+      }
+    },
+    "additionalProperties": false,
+    "required": ["id", "type", "for_list", "content", "repeating_blocks"]
```
katie-gardner, line 62: Was it agreed that repeating blocks would be mandatory?

katie-gardner (changes requested):
Beyond small comment json schema looks good - I think validator should have a valid test schema though

I'm also wondering if this invalidates the test in `app/validators/blocks/list_collector_validator.py` around single collectors with repeating blocks and if this perhaps needs to update the test to check against both types of list collector

schemas/blocks/list_collector_content.json
petechd: Ok, I've added conditional required `content` or `content_variant` to the rules now. Thanks.

schemas/blocks/list_collector_content.json
petechd: So here we made an assumption that repeating blocks will be always required? Since otherwise, there will be nothing to display after initial list collector content block?

katie-gardner: ok cool, makes sense - just checking 🙂

MebinAbraham: We don't need to enforce that this needs repeating blocks. It might have it. It allows them to use it as a summary preview. i.e Here are the list of things you need to report for. Then they might get a repeating section for each

MebinAbraham:
> I'm also wondering if this invalidates the test in `app/validators/blocks/list_collector_validator.py` around single collectors with repeating blocks and if this perhaps needs to update the test to check against both types of list collector

👍 Yes we should. `ListCollectorContent` needs to be added to the `blocks/__init__.py` file. The two should share common code.

Yuyuutsu (approved):
LGTM !

app/validators/blocks/list_collector_content_validator.py
```diff
@@ -0,0 +1,42 @@
...
+    NON_SINGLE_REPEATING_BLOCKS_LIST_COLLECTOR = (
+        "List may only have one List Collector and List Collector Content, if "
+        "the List Collector Content features Repeating Blocks"
+    )
+
+    def validate(self):
+        super().validate()
+        try:
+            self.validate_repeating_blocks_list_collector_content()
+        except KeyError as e:
```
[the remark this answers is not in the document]
petechd: Correct, no way this can key error on something that other validator wouldn't catch. Fixed.

app/validators/blocks/list_collector_validator.py
```diff
@@ -104,16 +109,24 @@ def validate_other_list_collectors(self):
                     list_name=list_name,
                 )
 
-    def validate_single_repeating_blocks_list_collector(self):
-        if self.block.get("repeating_blocks"):
-            list_name = self.block["for_list"]
 
-            other_list_collectors = self.questionnaire_schema.get_other_blocks(
-                self.block["id"], for_list=list_name, type="ListCollector"
+def validate_repeating_blocks_list_collectors(validator, same_type, other_type):
```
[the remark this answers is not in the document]
petechd: Ok, fixed inheritance and how we call this method so it looks better now, thanks.

katie-gardner (approved):
Nice one 🎉

app/validators/blocks/list_collector_validator.py
```diff
@@ -104,16 +109,23 @@ def validate_other_list_collectors(self):
...
+            )
+            other_list_collectors_same_type = (
+                self.questionnaire_schema.get_other_blocks(
+                    self.block["id"], for_list=list_name, type=same_type
+                )
             )
 
-            if other_list_collectors:
+            if (
+                list_collectors_other_type and len(list_collectors_other_type) > 1
```
katie-gardner, line 126: minor non-blocker `get_other_blocks` always returns a list, so this could just check `len(list_collectors_other_type) > 1`

katie-gardner: wait, actually, why does it need to check its bigger than 1? Is it OK to have a ListCollector and ListCollectorContent for the same list both with repeating blocks?

MebinAbraham: That should work fine.

katie-gardner: in that case - all good from me

app/validators/blocks/list_collector_validator.py
```diff
@@ -29,7 +29,10 @@ class ListCollectorValidator(BlockValidator, ValidateListCollectorQuestionsMixin
         "Multiple list collectors populate a list using different "
         "answer_ids in the add block"
     )
-    NON_SINGLE_REPEATING_BLOCKS_LIST_COLLECTOR = "List may only have one List Collector, if the List Collector features Repeating Blocks"
+    NON_SINGLE_REPEATING_BLOCKS_LIST_COLLECTOR = (
+        "List may only have one List Collector and List Collector Content, if "
+        "the List Collector Content or List Collector features Repeating Blocks"
+    )
```
MebinAbraham, line 35: I probably missed this when it went in, why would we not allow this?. We should only now allow it if it is within the same section. If two collectors reference the same list in two different sections, we should allow it.

katie-gardner: I thought it made sense when reading this so that if you had two "people" list collectors and with different repeating blocks, adding people on the second list would leave the first list incomplete, and you'd need to go back for those repeating blocks, so it seems user friendly for the two sets of repeating blocks to be on the same list collector? Don't see why it wouldn't work with multiple though

MebinAbraham: Ah I was thinking of it like, the additional list collector could only be off type content so they wouldn't allow adding.

katie-gardner: Ahh so limit to 1 normal list collector with repeating blocks but allow any number of content ones? don't see any harm in that

MebinAbraham: > though I interpreted the comment above from the first review to mean we might not want this? was the original validation what was wanted?

Not sure I am following.

The original seem to only check that we don't have more than 1.

I think it should be: only 1 per list name per section. However, any amount of additional are allowed if they are content list collectors.

katie-gardner: Ah of course, the existing probably misses the section limit for content ones, but can have multiple in different sections and stick to a single normal + repeating one

Makes sense to me now 👍 agree with your verdict

petechd: So, after reading all this, correct me if I'm wrong but my validation is checking now that **there is only 1 of either type per list name and per section**.

petechd: > We need to enforce only 1 list per section when section summary is used (this might exist already)

Ok, I've added this check in section summary validation part of section validation now.

app/validators/blocks/list_collector_validator.py
```diff
@@ -110,20 +110,24 @@ def validate_other_list_collectors(self):
...
-            )
-            other_list_collectors_same_type = (
-                self.questionnaire_schema.get_other_blocks(
-                    self.block["id"], for_list=list_name, type=same_type
-                )
+        if not self.block.get("repeating_blocks"):
+            return
+        list_name = self.block["for_list"]
+        section_id = self.questionnaire_schema.get_section_id_for_block(self.block)
+        list_collectors_other_type = self.questionnaire_schema.get_other_blocks(
```
katie-gardner, line 117: Since you're adding the results together, I think this can be simplified - remove the params `same_type` and `other_type` and then this could be 
`other_list_collectors = self.questionnaire_schema.get_other_blocks(self.block["id"], for_list=list_name)`
and use that in the if statement

app/validators/questionnaire_schema.py
```diff
@@ -536,3 +536,16 @@ def get_answers_from_question(question):
             *question.get("dynamic_answers", {}).get("answers", []),
             *question.get("answers", []),
         ]
+
+    def get_section_block_ids(self, current_section=None):
+        block_ids = []
+        section = self.get_section(current_section)
```
katie-gardner, line 542: Will this work if current section is None? is there a reason to call it current section instead of just section?

I think you can simplify this function to just return
```
self.blocks_by_section_id[section]
```

app/validators/questionnaire_schema.py
```diff
@@ -536,3 +536,16 @@ def get_answers_from_question(question):
...
+        for group in section.get("groups"):
+            block_ids.extend(block["id"] for block in group.get("blocks"))
+        return block_ids
+
+    def get_section_id_for_block(self, block):
+        for section in self.sections:
```
katie-gardner, line 548: possible simplification:
```
for section_id, blocks in self.blocks_by_section_id.items():
 if block in blocks:
  return section_id
```

app/validators/sections/section_validator.py
```diff
@@ -40,11 +40,28 @@ def validate_repeat(self):
             self.validate_list_exists(section_repeat["for_list"])
 
     def validate_summary(self):
-        section_summary = self.section.get("summary", None)
+        if not (section_summary := self.section.get("summary", None)):
```
katie-gardner, line 43: dont need the default to None

app/validators/sections/section_validator.py
```diff
@@ -40,11 +40,28 @@ def validate_repeat(self):
...
+            self.validate_list_exists(item.get("for_list"))
 
-        if section_summary:
-            for item in section_summary.get("items", []):
-                self.validate_list_exists(item.get("for_list"))
+        for_lists = []
+        content_block_in_section = False
+
+        for block_id in self.questionnaire_schema.get_section_block_ids(
+            self.section.get("id")
```
katie-gardner, line 52: `self.section["id"]`  instead?

app/validators/sections/section_validator.py
```diff
@@ -40,11 +40,28 @@ def validate_repeat(self):
...
+        ):
+            block = self.questionnaire_schema.get_block(block_id)
+            if (
+                block.get("type") in ["ListCollector", "ListCollectorContent"]
+                and block.get("for_list") not in for_lists
```
katie-gardner, line 57: this should exist if its a collector type so would change to `block["for_list"]`

app/validators/sections/section_validator.py
```diff
@@ -40,11 +40,28 @@ def validate_repeat(self):
...
-                self.validate_list_exists(item.get("for_list"))
+        for_lists = []
+        content_block_in_section = False
+
+        for block_id in self.questionnaire_schema.get_section_block_ids(
+            self.section.get("id")
+        ):
+            block = self.questionnaire_schema.get_block(block_id)
+            if (
+                block.get("type") in ["ListCollector", "ListCollectorContent"]
```
katie-gardner, line 56: blocks should have types so `block["type"]` seems more suitable

app/validators/sections/section_validator.py
```diff
@@ -40,11 +40,28 @@ def validate_repeat(self):
...
+                and block.get("for_list") not in for_lists
+            ):
+                for_lists.append(block.get("for_list"))
+            if block.get("type") == "ListCollectorContent":
```
katie-gardner, line 60: same for these two lines, dont think .get needed

app/validators/sections/section_validator.py
```diff
@@ -40,11 +40,28 @@ def validate_repeat(self):
             self.validate_list_exists(section_repeat["for_list"])
 
     def validate_summary(self):
```
katie-gardner, line 42: I'm getting slightly confused about what this is doing, I think a docstring for this one would be helpful

app/validators/sections/section_validator.py
```diff
@@ -40,11 +40,28 @@ def validate_repeat(self):
...
+            self.section.get("id")
+        ):
+            block = self.questionnaire_schema.get_block(block_id)
+            if (
+                block.get("type") in ["ListCollector", "ListCollectorContent"]
+                and block.get("for_list") not in for_lists
+            ):
+                for_lists.append(block.get("for_list"))
+            if block.get("type") == "ListCollectorContent":
+                content_block_in_section = True
```
katie-gardner, line 61: I'm not sure why the `content_block_in_section` variable is required? I would have thought multiple for lists with multiple list collectors is a problem either way?

app/error_messages.py
```diff
@@ -2,6 +2,7 @@
 INVALID_WHITESPACE_FOUND = "Found invalid white space(s) in schema text"
 DUPLICATE_ID_FOUND = "Duplicate id found"
 FOR_LIST_NEVER_POPULATED = "for_list is not populated by any ListCollector blocks"
+MULTIPLE_FOR_LISTS = "Multiple for_lists in a section when section summary enabled"
```
katie-gardner, line 5: does this get tested anywhere?

app/validators/blocks/list_collector_validator.py
katie-gardner: I've gotten a bit lost, I think its worth adding a clear docstring wherever this is checked
I think the above alludes to having:

a given list to only have one list collector within a given section - (regardless of type I think?)

a list to only have one list collector with repeating blocks, if there are other list collector content blocks with repeating blocks that use it?
_____
So if I have a list collector of employees with some repeating blocks LC1
and a content block LCC1 with some more repeating questions, and another content block LCC2 with some more questions - I think that this is ok?

but if I had LC1 and then could also add to the list in LC2 with those questions, I think that would not be fine
_____
but im not sure thats exactly what is verified

katie-gardner (changes requested):
Slightly unsure on the exact scenarios that are allowed/not, few comments anyway and some extra comments would help out

app/validators/blocks/list_collector_validator.py
```diff
@@ -104,16 +109,25 @@ def validate_other_list_collectors(self):
...
+        other_list_collectors_same_type = self.questionnaire_schema.get_other_blocks(
+            self.block["id"], for_list=list_name, type=same_type
+        )
+        if [
+            list_collector
+            for list_collector in list_collectors_other_type
+            + other_list_collectors_same_type
+            if list_collector["id"]
+            in self.questionnaire_schema.get_section_block_ids(section_id)
+        ]:
```
MebinAbraham, line 129: I find this whole thing pretty hard to read. This definitely need comments.

Why do we get same type and other type?. It is always passing two?.

app/validators/blocks/list_collector_validator.py
katie-gardner: To clarify - this is checking that if you find any other list collectors in the section for the same list, thats an error because the section should only have 1?
Which sounds correct to me - just wanted to check

app/validators/blocks/list_collector_validator.py
MebinAbraham: If we have a list name used twice but in different sections, why would we not allow it?.

katie-gardner: I think the idea behind not allowing it was if both of them were normal list collectors with repeating blocks
so if you had people LC + questions about bills (section 1)
then people LC + questions about dependents (section 2)
if you added a person in section 2, the repeating blocks for section 1 are then not complete, so you have to go back

which might be fine / an author choice, but I thought that slightly undesirable route might be why we wouldn't want multiple normal ones with repeating blocks for the same list

app/error_messages.py
```diff
@@ -2,7 +2,9 @@
 INVALID_WHITESPACE_FOUND = "Found invalid white space(s) in schema text"
 DUPLICATE_ID_FOUND = "Duplicate id found"
 FOR_LIST_NEVER_POPULATED = "for_list is not populated by any ListCollector blocks"
-MULTIPLE_FOR_LISTS = "Multiple for_lists in a section when section summary enabled"
+MULTIPLE_LIST_COLLECTORS_WITH_LIST_SUMMARY_ENABLED = (
+    "Multiple list collectors for same list in a section when list summary enabled"
+)
```
MebinAbraham, line 7: I don't think this message is very clear. Maybe `Section cannot have multiple list collectors that populate the same list when summary with items are enabled.`

I think we need to do a piece to update all validator error messages to ensure they describe what has happened and how to fix it.

app/error_messages.py
```diff
@@ -2,7 +2,9 @@
 INVALID_WHITESPACE_FOUND = "Found invalid white space(s) in schema text"
 DUPLICATE_ID_FOUND = "Duplicate id found"
 FOR_LIST_NEVER_POPULATED = "for_list is not populated by any ListCollector blocks"
-MULTIPLE_FOR_LISTS = "Multiple for_lists in a section when section summary enabled"
```
MebinAbraham: Need to capture this on the card to ensure we do support this.

app/validators/blocks/list_collector_validator.py
petechd: It's removed now since we want to check for multiple list collector on section level, so I reverted it back to its original state when introduced by Kyle.

app/validators/blocks/list_collector_validator.py
petechd: It's removed now since we want to check for multiple list collector on section level, so I reverted it back to its original state when introduced by Kyle.

app/validators/sections/section_validator.py
petechd: This code is removed now.

app/validators/sections/section_validator.py
petechd: Added a docstring.

app/validators/blocks/list_collector_validator.py
petechd: As we discussed,  added new multiple list collectors validation on section level.

app/validators/sections/section_validator.py
```diff
@@ -360,3 +366,23 @@ def _validate_answers(self, question):
...
+
+    def _validate_multiple_list_collectors(self):
+        for_lists = []
+
+        for block_id in self.questionnaire_schema.get_section_block_ids(
+            self.section["id"]
+        ):
+            block = self.questionnaire_schema.get_block(block_id)
+            if block["type"] in ["ListCollector", "ListCollectorContent"]:
+                if block.get("summary"):
```
katie-gardner, line 378: why do you break if the block has a summary?

katie-gardner (changes requested):
If I amend the invalid schema, and delete the list collector content, but add another list collector for people, and then delete `show_non_item_answers` from the summary, the schema passes validation, but I get an internal error on trying to load the summary page. This hasn't fully prevented list collectors for different lists being disallowed when summary has items

app/validators/sections/section_validator.py
petechd: This summary field present means section will have a list on section summary ("list summary") enabled, and we want to allow multiple list collectors when such "list summary" is enabled. From what @MebinAbraham said: "if summary without items or no summary or it is a list summary (census version) then allow multiple else disallow".

app/validators/sections/section_validator.py
```diff
@@ -360,3 +366,23 @@ def _validate_answers(self, question):
...
+                    break
+
+                if block["for_list"] not in for_lists:
```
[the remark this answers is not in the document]
petechd: Yeah, from what we understand, it doesn't matter if they are different lists.

app/validators/sections/section_validator.py
katie-gardner: ohhh, I thought it was list summary or summary with items that was banned but that makes sense, I see!

If you had one list with a summary, and one without, but the first one breaks out of the loop, could that break this?

petechd:
> If I amend the invalid schema, and delete the list collector content, but add another list collector for people, and then delete `show_non_item_answers` from the summary, the schema passes validation.
> 
> I'm not certain if this is this should be allowed - it does seem to work, and only the items as setup in the section summary config show on the summary page - was this intentionally enabled?

Is this because the first collector is for companies and the second for people list? I tried this exact scenario for LC with same for list and it errors where it's supposed to.

app/validators/sections/section_validator.py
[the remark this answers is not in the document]
katie-gardner: I'd add a comment to mention that just so its clear its intentionally allowed

app/validators/blocks/list_collector_validator.py
```diff
@@ -105,15 +105,14 @@ def validate_other_list_collectors(self):
...
 
     def validate_single_repeating_blocks_list_collector(self):
-        if self.block.get("repeating_blocks"):
-            list_name = self.block["for_list"]
-
-            other_list_collectors = self.questionnaire_schema.get_other_blocks(
-                self.block["id"], for_list=list_name, type="ListCollector"
+        if not self.block.get("repeating_blocks"):
+            return
+        list_name = self.block["for_list"]
```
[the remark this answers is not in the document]
katie-gardner: Yeah im happy with that - don't want to block this one 👍

app/validators/sections/section_validator.py
katie-gardner: I think it would be helpful to have that explanation in a comment - that validation isn't needed for list summaries, and since the possible bug mentioned is existing before this PR, that could be a follow up too

app/validators/sections/section_validator.py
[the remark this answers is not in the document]
petechd: Comments added, thanks!

katie-gardner (approved):
LCC validation works as expected 🎉 happy for multiple list collector validation to be in the follow up

rmccar (approved):

PR #206, opened 2023-07-05
Looping 4 - Allow calculated summary using single repeating answer

katie-gardner:
This PR updates existing validation to calculated summaries to allow a calculated summary using a single answer id from a repeating block

One of the existing checks is that any answer in a calculated summary comes before the calculated summary. This check was not working correctly because we were populating the questionnaire schemas block ids, first with standard blocks, then add/edit/remove and then repeating blocks, resulting in repeating blocks always coming after the calculated summary blocks

To fix this a method has been added that converts the json path of each block match, into a tuple of its position, and this is used to sort the blocks

The only other change needed was to update the existing check in `calculated_summary_block_validator` to allow the single answer id to correspond to dynamic _or_ repeating question answers.

[Corresponding runner PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1157)

### How to review
Check that with the new changes to `questionnaire_schema` that the blocks are now showing in the correct order. Additionally confirm that this allows a calculated summary with a single repeating block answer id or dynamic answer id, but not a single standard answer. This shouldn't need any additional tests but comment if you think of one

### Checklist

### DIFF
--- a/app/validators/blocks/calculated_summary_block_validator.py
+++ b/app/validators/blocks/calculated_summary_block_validator.py
@@ -6,7 +6,7 @@ class CalculatedSummaryBlockValidator(CalculationBlockValidator):
         "Answer ids for calculated summary must be set before calculated summary block"
     )
     ANSWER_SET_IN_DIFFERENT_SECTION_FOR_CALCULATED_SUMMARY = "Answer ids for calculated summary must be set in the same section as the calculated summary block"
-    CALCULATED_SUMMARY_WITH_NON_DYNAMIC_SINGLE_ANSWER = "Calculated summaries cannot consist of a single answer unless it is a dynamic answer"
+    CALCULATED_SUMMARY_WITH_NON_REPEATING_SINGLE_ANSWER = "Calculated summaries cannot consist of a single answer unless it is a repeating answer"
 
     def __init__(self, block, questionnaire_schema):
         super().__init__(block, questionnaire_schema)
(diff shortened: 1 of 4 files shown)


app/validators/blocks/calculated_summary_block_validator.py
```diff
@@ -20,31 +20,40 @@ def validate(self):
...
-                self.add_error(
-                    self.CALCULATED_SUMMARY_WITH_NON_DYNAMIC_SINGLE_ANSWER,
-                    block_id=self.block["id"],
-                    answer_id=single_answer_id,
-                )
+                return
+            # check if it's for a repeating question
+            question_block = self.questionnaire_schema.get_block_by_answer_id(
+                single_answer_id
+            )
```
MebinAbraham, line 48: Minor and non blocker, assign this upfront given it can be used twice, if the first if is false.

app/validators/blocks/calculated_summary_block_validator.py
```diff
@@ -20,31 +20,40 @@ def validate(self):
...
+            if question_block["type"] == "ListRepeatingQuestion":
+                return
+
+            self.add_error(
+                self.CALCULATED_SUMMARY_WITH_NON_DYNAMIC_SINGLE_ANSWER,
```
MebinAbraham, line 53: Worth changing the var name from dynamic to repeating answer then?.

app/validators/questionnaire_schema.py
```diff
@@ -92,16 +103,17 @@ def get_context_from_match(match):
...
-        }
+        self.matches = [
+            *parse("$..blocks[*]").find(self.schema),
+            *parse("$..[add_block, edit_block, add_or_edit_block, remove_block]").find(
+                self.schema
+            ),
+            *parse("$..repeating_blocks[*]").find(self.schema),
+        ]
+        # order the blocks by the order that they occur in the json schema
+        self.sorted_matches = sorted(self.matches, key=json_path_position)
```
MebinAbraham, line 114: 👍

MebinAbraham (commented):
Minor comments otherwise LGTM.

app/validators/blocks/calculated_summary_block_validator.py
katie-gardner: Good shout - done 👍

app/validators/blocks/calculated_summary_block_validator.py
katie-gardner: Yep, agree 👍

MebinAbraham (approved):

MebinAbraham:
Hold off merging this until the Runner PR is ready just in case something pops up.

Yuyuutsu (approved):

PR #209, opened 2023-07-20
Add support for supplementary data as a value source

katie-gardner:
This PR adds support for Supplementary Data to be used as a value source in schemas. It supports value sources with a `source` of `"supplementary_data"`, an identifier corresponding to the top level key of the data being referenced, and an optional list of selectors for accessing nested data.

It is added in all places that allow any value source/ any non-location value source and also a a source for calculations, provided the value is numeric

The changes are covered by AJV validation

Functional use of the schema change is shown in the [corresponding runner PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1169)

### How to review
Check that the changes to the value source schema make sense

### Checklist


schemas/value_sources.json
```diff
@@ -199,5 +209,28 @@
...
+        "description": "The top level key within `data` for the source of the data"
+      },
+      "source": {
+        "type": "string",
+        "enum": ["supplementary_data"]
+      },
+      "selectors": {
+        "description": "A list of selectors for accessing data within a nested structure",
+        "type": "array",
+        "items": {
```
MebinAbraham, line 228: We should have `minItems` set to 1.

schemas/value_sources.json
```diff
@@ -199,5 +209,28 @@
...
+    "description": "A reference to a value from supplementary data. When resolved this can be a number, string, array or null.",
+    "type": "object",
+    "properties": {
+      "identifier": {
+        "$ref": "common_definitions.json#/non_empty_string",
+        "description": "The top level key within `data` for the source of the data"
+      },
+      "source": {
+        "type": "string",
+        "enum": ["supplementary_data"]
```
MebinAbraham, line 223: Non blocker. existing pattern but we should start using `const` when there is only one value. `const` didn't exist back in the day. Maybe a separate card to enforce across the board.

MebinAbraham (commented):
We should update some of the files in `tests/schemas/rules/`. Both the `values_sources.json` and `value.json` as a minimum.

schemas/value_sources.json
katie-gardner: Fixed this one and raised a card for the others 👍

MebinAbraham: Nice one.

MebinAbraham (approved):

kylelawsonAND (approved):

PR #210, opened 2023-07-25
Upgrade to Python 3.11

rmccar:
Upgrade to Python 3.11

### How to review
- Tests pass
- Changes make sense

### Checklist


rmccar:
[Upgrade to Python 3.11 on non Runner repos (2 SP)](https://trello.com/c/5rkGy42V/5730-upgrade-to-python-311-on-non-runner-repos-2-sp)

katie-gardner (changes requested):
I'm getting warnings that the `poetry.lock` file is not consistent with the toml file, think it needs to be regenerated (it suggests using `--no-update` to avoid all the packages updating at the same time)

MebinAbraham:
> I'm getting warnings that the `poetry.lock` file is not consistent with the toml file, think it needs to be regenerated (it suggests using `--no-update` to avoid all the packages updating at the same time)

It's worth us updating the packing at the same time.

katie-gardner (approved):
All working post upgrade and matches my generated poetry file 👍

Yuyuutsu (commented):
Validator working as expected, and changes look good. I am also getting the 
```
Installing dependencies from lock file
Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies. Run `poetry lock [--no-update]` to fix it.
```
if that is an issue

pyproject.toml
```diff
@@ -8,15 +8,15 @@ description = "API for questionnaire schema validation"
...
+python = "^3.11"
 Flask = "^2.0.2"
 jsonpointer = "^2.2"
 structlog = "^21.5.0"
 python-dateutil = "^2.8.2"
 jsonschema = "4.4.0"
 jsonpath-rw = "^1.4.0"
 jsonpath-rw-ext = "^1.2.2"
-eq-translations = {git = "https://github.com/ONSDigital/eq-translations.git", rev = "v4.9.0"}
+eq-translations = {git = "https://github.com/ONSDigital/eq-translations.git", rev = "v4.9.1"}
```
MebinAbraham, line 19: While we are here, we should update these versions to the latest available. Useful plugin: https://github.com/MousaZeidBaker/poetry-plugin-up

MebinAbraham: We should update to latest available, not latest within the version constraints.

rmccar: Pretty sure thats what I have done, I used `poetry up --latest`

MebinAbraham: Right, I think I may have been looking at the first commits which wasn't. All good then 👍

rmccar: Ah yeah, I ran it without `--latest` at first 👍

MebinAbraham (approved):

PR #213, opened 2023-08-31
Support new supplementary lists property

katie-gardner:
This PR is the validator component of allowing a list collector content block/repeating section/ dynamic answers to use a `for_list` that refers to a list that is not in the schema, provided it is in a top level property `supplementary_data.lists` within the schema, implying that the list will be populated by supplementary data.

Since `ListCollectorContentBlock` now needs validation, a validator has been added for this, and tests to make sure that `for_list` always comes from either a list collector or supplementary data. It also validates that you cannot use a normal list collector with a list defined in this top level property, as supplementary lists should not be modified.

The Runner validation to check the required lists are actually passed in is [this corresponding PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1198)

### How to review
Check the test coverage is suitable and it doesn't miss any edge cases where lists may be undefined (The checking that the supplementary data list is actually passed in has to be in runner). Check that the change to the questionnaire schema is as intended.

### Checklist


katie-gardner:
[Support for repeating sections / list collector content blocks without a list collector (2 SP)](https://trello.com/c/cyFoALKE/5786-support-for-repeating-sections-list-collector-content-blocks-without-a-list-collector-2-sp)

app/validators/blocks/list_collector_validator.py
```diff
@@ -21,14 +21,11 @@ class ListCollectorValidator(BlockValidator, ValidateListCollectorQuestionsMixin
...
     )
-    LIST_COLLECTOR_ADD_EDIT_IDS_DONT_MATCH = (
-        "The list collector block contains an add block and edit block"
-        " with different answer ids"
-    )
-    NON_UNIQUE_ANSWER_ID_FOR_LIST_COLLECTOR_ADD = (
-        "Multiple list collectors populate a list using different "
-        "answer_ids in the add block"
+    LIST_COLLECTOR_FOR_SUPPLEMENTARY_LIST_IS_INVALID = (
+        "List collectors cannot be for a list which comes from supplementary data"
```
MebinAbraham, line 25: For clarity, this should be either `Non content list collectors ...` or `Only ListCollectorContent types can be used for a list which comes from supplementary data`

app/validators/blocks/list_collector_content_validator.py
```diff
@@ -0,0 +1,22 @@
+from app.error_messages import FOR_LIST_NEVER_POPULATED
+from app.validators.blocks.block_validator import BlockValidator
+
+
+class ListCollectorContentValidator(BlockValidator):
```
MebinAbraham, line 5: Might be an existing issue but is `validate_single_repeating_blocks_list_collector` on ListCollectorValidator correct?. It doesn't account for `ListCollectorContent` but either way the validation would need to happen if the initial block was content block hence needs to be done here?

katie-gardner: I thought we only needed that validation for list collectors? So we ensure only one list collector with repeating blocks exists for list=people or something but allow multiple for list collector content as its possible you'd want to do that

that said, I think this would block having a conditionally chosen list collector with repeating blocks for a list, so we may still have an issue

MebinAbraham: Yes, I think you are right. In that case, all good.

> that said, I think this would block having a conditionally chosen list collector with repeating blocks for a list, so we may still have an issue

We'll revisit if it becomes an issue

MebinAbraham (approved):

MebinAbraham:
Merges once runner is ready to prevent other PRs being blocked as a result of this change.

PR #214, opened 2023-09-07
Add validation of numeric answer types in calculations 

petechd:
Adds validation of numeric answer types used in calculated question validator. Previously, calculations allowed different types of answers eg. calculation of currency type answers accepts percentage type answer to be validated against.

### How to review
Run new invalid schema through validator, different answer types should now be detected and new error message with detailed list of answer types should be displayed. This should work for both `answer_id` in the calculation, and value sources referring to an answer.

### Checklist


tests/schemas/invalid/test_invalid_validation_sum_against_total_different_answer_types.json
```diff
@@ -0,0 +1,117 @@
+{
+  "mime_type": "application/json/ons/eq",
+  "language": "en",
+  "schema_version": "0.0.1",
+  "data_version": "0.0.3",
+  "survey_id": "0",
+  "title": "A test schema to validate a sum of answers are Equal to a given total",
```
katie-gardner, line 7: Need a test that asserts the errors thrown out by this schema

app/validators/questions/calculated_question_validator.py
```diff
@@ -55,10 +56,27 @@ def validate_calculations_value_source_is_numeric(self):
...
 
                 if value.get("source") == "answers":
                     self._validate_answer_is_numeric(answer_id)
+
+    def _validate_answers_are_same_numeric_type(self, answers_to_calculate, answer_id):
+        types = [self.schema.get_answer_type(answer_id)]
+        types.extend(
+            self.schema.get_answer_type(answer) for answer in answers_to_calculate
+        )
+        types_set = set(types)
```
katie-gardner, line 75: this could be simplified with a set comprehension
```
types = {
 self.schema.get_answer_type(answer) for answer in [*answers_to_calculate, answer_id]
}
```

katie-gardner (changes requested):

app/validators/questions/calculated_question_validator.py
```diff
@@ -46,19 +50,35 @@ def _validate_answer_is_numeric(self, answer_id):
...
+                # Calculated summary value source is validated elsewhere and must be of a number type
+                if isinstance(value, dict) and value.get("source") == "answers":
+                    answer_id = value.get("identifier")
 
-            if answer_id := calculation.get("answer_id"):
+            if answer_id:
                 self._validate_answer_is_numeric(answer_id)
+                self._validate_answers_are_same_numeric_type(
+                    calculation.get("answers_to_calculate"), answer_id
+                )
```
MebinAbraham, line 68: This would mean `answers_to_calculate` isn't checked if `answer_id` isn't set.

We should move this outside the if and handle answer id not existing in the validate method

katie-gardner: Good spot - and we'd also have an issue if they aren't numeric and no referenced `answer_id`, moved both of these checks out of the if and passed them a dict of answer types instead + new test for this scenario

Yuyuutsu (approved):
Functionally tested and changes looks good to me 👍

PR #219, opened 2023-10-06
Replace instances of enum using a single value with const

berroar:
Uses const instead of enum where there is only one single value.

### How to review
Check no instances have been missed

### Checklist


berroar:
[Use const instead of enum for single values in validator (1 SP)](https://trello.com/c/thLmFGqk/5805-use-const-instead-of-enum-for-single-values-in-validator-1-sp)

ajv/tests/test_questionnaire_validator.js
```diff
@@ -38,10 +38,10 @@ describe("AJV schema Validator", () => {
           res.should.have.status(200);
           const failure = res.body.errors.pop();
           res.body.should.have.property("success", false);
-          failure.should.have.property("keyword", "enum");
+          failure.should.have.property("keyword", "const");
```
katie-gardner, line 41: I know this was already there, but I think it would be worth changing the name of this schema to `test_invalid_block_type.json` because the error is due to the block having `"type": "Test"` as opposed to `"type": "Question"` which I wasn't expecting based off the name

katie-gardner (approved):
Looks good - no other cases of enums with multiple values and all changes correct 👍 
One minor suggestion

ajv/tests/test_questionnaire_validator.js
berroar: Updated 👍

rmccar (approved):

PR #227, opened 2023-11-28
Prevent repeating GCS referencing static CS with answers repeating for same list

katie-gardner:
This PR prevents a repeating Grand Calculated Summary from including a static Calculated Summary that has repeating answers for the same list. It is in place to support the [corresponding runner PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1268)

This included some changes to `questionnaire_schema` to include a way of looking up a list name by answer id. I wasn't aware of another way to easily do this.
 
### How to review
Check over the added invalid test schema and make sure it is correctly covering the case that the card mentions to exclude. GCS should allow a static CS repeating over a different list, but not the same one. Look for any possible edge cases this has missed.

### Checklist


katie-gardner:
[Grand Calculated Summary in repeat: Support for non-repeating Calculated Summaries with repeating answers (5 SP)](https://trello.com/c/s7JzRKrJ/5892-grand-calculated-summary-in-repeat-support-for-non-repeating-calculated-summaries-with-repeating-answers-5-sp)

app/validators/questionnaire_schema.py
```diff
@@ -143,6 +143,19 @@ def __init__(self, schema):
             '$..blocks[?(@.type=="ListCollector")].for_list', self.schema
         )
         self.list_names = self.list_collector_names + self.supplementary_lists
+        self.list_names_by_repeating_block_id = {
```
petechd, line 146: Maybe it's just me but two for loops in this comprehension, I would do just two normal loops for readability's sake. Same  the next one below, but up to you and other devs.

app/validators/questionnaire_schema.py
```diff
@@ -396,6 +409,19 @@ def get_all_dynamic_answer_ids(self, block_id):
             for answer in question.get("dynamic_answers", {}).get("answers", [])
         }
 
+    def get_list_name_for_answer_id(self, answer_id: str) -> str | None:
+        """If the answer is repeating, return the name of the list it repeats over, otherwise None"""
```
petechd, line 413: If the answer is repeating or block type is list collector? Or something like `if answer is repeating, dynamic, part of list collector or repeating section`?

katie-gardner: Sure, changed it to "If the answer is dynamic or in a repeating block or section, return the name of the list it repeats over otherwise None" to match what runner says

app/validators/questionnaire_schema.py
katie-gardner: Yeah I think the second of these two properties is probably a bit much, I've changed it to a cached property with a normal for loop. For the first one, I've made list collectors a shared variable so now its still two loops but I think easier to follow, let me know what you think? can always do another cached property if preferable

app/validators/blocks/grand_calculated_summary_block_validator.py
```diff
@@ -83,39 +87,80 @@ def validate_calculated_summary_is_before_grand_calculated_summary_block(
...
         """
-        If the grand calculated summary references a repeating calculated summary, this is only valid if:
-        1) the grand calculated summary is also repeating
-        2) it is in the same repeating section as the repeating calculated summary it references
+        Run additional validation for the scenarios:
+        1) any grand calculated summary referencing a repeating calculated summary
+        2) repeating grand calculated summary referencing a static calculated summary
         """
-        gcs_section_id = self.questionnaire_schema.get_section_id_for_block_id(
+        gcs_section = self.questionnaire_schema.get_parent_section_for_block(
```
berroar, line 96: Only minor, but quite a mixture of using cs/calculated_summary and gcs/grand_calculated_summary in these files, would be good to make this consistent.

katie-gardner: Yeah the variable names were just really long, but consistency is important I agree, have made them all full name

app/validators/questionnaire_schema.py
petechd: Thanks, much easier to follow now.

petechd (approved):

tests/validators/blocks/test_grand_calculated_summary_block_validator.py
```diff
@@ -125,3 +125,38 @@ def test_invalid_non_repeating_grand_calculated_summary_referencing_repeating_ca
     errors = validator.validate()
 
     assert errors == expected_error_messages
+
+
+def test_invalid_repeating_grand_calculated_summary_with_static_calculated_summary_with_repeating_answers_same_list():
```
petechd, line 130: Not anything blocking and just a detail (I can't find much otherwise, the code looks good) but it is the longest test title I've seen, it even breaks the Python's standard 79 characters a line rule so maybe use abbreviations for gcs and cs?

katie-gardner: I've got other review comments asking for consistency between abbreviations and full names, so I want to keep it consistent but I have reworded it slightly to make it shorter since the doc string makes it clear anyway

MebinAbraham (approved):

PR #228, opened 2023-12-05
Prevent duplicate answer_ids across different list collectors

liamtoozer:
An answer ID is globally unique in a schema, except for list collectors with the same name when used in one or more locations - it is correct that validation allows for this. 

However, when there are multiple list collectors with a **different** name in a schema, the answer ID needs to be unique for each list collector child block (add, edit, and remove). Validation needs to be in place to ensure this.

The consequence of not validating this is that when Runner loads a schema with 2 list collectors, but both share an `answer_id` in the `add_block`, `edit_block`, or `remove_block`, if there is a routing rule in place to skip the first list collector then **both** list collectors are removed from the path, instead of just the first.

Additionally, we can also validate to make sure that a list collector child block answer ID is not used **anywhere else** in the schema if it's not for the same list.

Changes for this PR:
* I've updated the `validate_other_list_collectors` to include _all_ list collectors in a schema so that we can also validate against those. 
* As mentioned in the [PR comment by Katie,](https://github.com/ONSdigital/eq-questionnaire-validator/pull/228#pullrequestreview-1775633331) we can also validate to ensure that the answer IDs in child blocks for every list collector are unique for the answer_ids across the rest of the schema

### How to review
To review, try to validate the schema linked in the Trello ticket - it should now show the new validation error message. 

Would appreciate any advice/feedback on improvements too!


liamtoozer:
[Bug: Validator allows duplicate answer ids in different list collectors](https://trello.com/c/fB5R54hy/5965-bug-validator-allows-duplicate-answer-ids-in-different-list-collectors)

katie-gardner (changes requested):
~~Only functionally tested so far, but~~
- ~~if I change the invalid example added in this PR so that the visitor LC has `visitor-first-name` I get no errors 👍~~
- ~~If I then change the edit block to `edit-visitor-first-name` I get "Different list collectors populate a list using duplicate answer_ids in the add block" However the error is actually that the edit block and add block don't match, not that different LCs have duplicates, I think something has gone wrong?~~

katie-gardner (changes requested):
Ignore above, it does correctly pick up add/edit not matching 👍 
There is a case that has slipped through, add/edit don't allow duplicates in different LCs now but two different list collectors could have the same answer-id in the remove block, might need small extension for this case

katie-gardner (changes requested):
Just a few minor comments 👍 
Also this doesn't prevent having 
block-1 with answer-1
list-collector with add-block with answer-1

Might not need fixing in this card? but it is a case that could get through

app/validators/blocks/list_collector_validator.py
```diff
@@ -90,26 +91,61 @@ def validate_not_for_supplementary_list(self):
...
         )
 
         for other_list_collector in other_list_collectors:
-            other_add_ids = self.questionnaire_schema.get_all_answer_ids(
-                other_list_collector["add_block"]["id"]
-            )
-            difference = add_answer_ids.symmetric_difference(other_add_ids)
-            if difference:
+            other_list_name = other_list_collector["for_list"]
+            are_list_names_matching = other_list_name == list_name
```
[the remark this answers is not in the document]
liamtoozer: That's a lot easier to follow, thanks 👍

app/validators/blocks/list_collector_validator.py
```diff
@@ -90,26 +91,61 @@ def validate_not_for_supplementary_list(self):
...
                     list_name=list_name,
                 )
 
+            # Validate other different list collector
+            if not are_list_names_matching:
+                for list_block_to_validate in list_blocks_to_validate:
+                    has_duplicate_answer_ids = any(
+                        answer_id
+                        in other_list_blocks_answer_ids[list_block_to_validate]
+                        for answer_id in list_blocks_answer_ids[list_block_to_validate]
```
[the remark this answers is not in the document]
liamtoozer: Ah that's cool! Also easier to read with those variable namings too. Thanks 👍

app/validators/blocks/list_collector_validator.py
```diff
@@ -23,7 +23,8 @@ class ListCollectorValidator(BlockValidator, ValidateListCollectorQuestionsMixin
     )
     LIST_COLLECTOR_FOR_SUPPLEMENTARY_LIST_IS_INVALID = "Non content list collectors cannot be for a list which comes from supplementary data"
     LIST_COLLECTOR_ADD_EDIT_IDS_DONT_MATCH = "The list collector block contains an add block and edit block with different answer ids"
-    NON_UNIQUE_ANSWER_ID_FOR_LIST_COLLECTOR_ADD = "Multiple list collectors populate a list using different answer_ids in the add block"
+    NON_UNIQUE_ANSWER_ID_FOR_SAME_LIST_COLLECTOR_ADD_BLOCK = "Multiple list collectors with same name populate a list using different answer_ids in add block"
```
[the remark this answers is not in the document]
liamtoozer: Yeah, that's more clear! thanks

app/validators/blocks/list_collector_validator.py
```diff
@@ -96,53 +111,45 @@ def validate_other_list_collectors(self):
...
+                    self.add_error(
+                        self.DIFFERENT_LIST_COLLECTOR_ADD_BLOCKS_FOR_SAME_LIST,
+                        list_name=list_name,
                     )
-                    if has_duplicate_answer_ids:
+            else:
+                for child_block in self.LIST_CHILD_BLOCKS_TO_VALIDATE:
+                    if (
+                        other_list_blocks_answer_ids[child_block]
+                        & list_blocks_answer_ids[child_block]
```
katie-gardner, line 146: Non blocker - since you used `.intersection` above instead of `&` I'd probably also do so here just for consistency

app/validators/blocks/list_collector_validator.py
```diff
@@ -67,19 +69,32 @@ def validate_list_collector_answer_ids(self, block):
...
         - Enforce the same answer_ids on add and edit sub-blocks
+        - Ensure that that child block answer_ids are not used elsewhere in the schema that's not another list collector
         """
-        add_answer_ids = self.questionnaire_schema.get_all_answer_ids(
-            block["add_block"]["id"]
-        )
-        edit_answer_ids = self.questionnaire_schema.get_all_answer_ids(
-            block["edit_block"]["id"]
-        )
+        list_block_answer_ids = {
```
katie-gardner, line 74: Now that you have 3 dict comprehensions exactly like this, might be worth a shared method like `get_list_block_answer_ids(block_id: str)`

app/validators/blocks/list_collector_validator.py
```diff
@@ -90,25 +106,52 @@ def validate_not_for_supplementary_list(self):
...
+                for child_block in self.LIST_CHILD_BLOCKS_TO_VALIDATE
+            }
+
+            if list_name == other_list_collector["for_list"]:
+                if list_blocks_answer_ids["add_block"].symmetric_difference(
+                    other_list_blocks_answer_ids["add_block"]
+                ):
+                    self.add_error(
+                        self.DIFFERENT_LIST_COLLECTOR_ADD_BLOCKS_FOR_SAME_LIST,
+                        list_name=list_name,
```
katie-gardner, line 140: I quite like that you have block id and other list collector block id in the error added below, adding that to this one too could be helpful

katie-gardner (approved):
Couple tiny non blockers, but all working for me now 👍

MebinAbraham:
> Just a few minor comments 👍 Also this doesn't prevent having block-1 with answer-1 list-collector with add-block with answer-1
> 
> Might not need fixing in this card? but it is a case that could get through

We should handle this given we are in this space. I think we should do it as part of this card but happy for it to be a seperate PR.

liamtoozer:
> > Just a few minor comments 👍 Also this doesn't prevent having block-1 with answer-1 list-collector with add-block with answer-1
> > Might not need fixing in this card? but it is a case that could get through
> 
> We should handle this given we are in this space. Happy for it to be a follow-up card to a separate PR. @liamtoozer can you raise a card or PR to fix this please but progress it next.

Just to confirm - this should hopefully be covered in this PR now. Updated the PR description to reflect this

app/validators/blocks/list_collector_validator.py
liamtoozer: Cool that makes sense, thanks

app/validators/blocks/list_collector_validator.py
liamtoozer: Great shout - thanks. Have created a common method for this in questionnaire_schema as `get_list_collector_answer_ids_by_child_block` to handle these 3 areas 👍

app/validators/blocks/list_collector_validator.py
liamtoozer: Good idea, thanks

berroar (approved):
Tested, LGTM 👍

PR #229, opened 2023-12-05
Fix Grand Calculated Summary Value Source Type

katie-gardner:
During development of [GCS Dependencies](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1270) it was noticed that validator doesn't correctly deduce the type for grand calculated summary value sources. This is simply because it was missed the first time and no schema had used a GCS in a when rule against a value. This PR has added the small fix to amend that.

### How to review
Make sure the test schema added (copied from runner) uses the GCS value source as you would expect and that the approach to get the type is working correctly.

### Checklist


app/validators/routing/types.py
```diff
@@ -60,6 +60,11 @@ def resolve_value_source_json_type(value_source, schema):
         answer_type = schema.answers_with_context[answer_id]["answer"]["type"]
         return ANSWER_TYPE_TO_JSON_TYPE[answer_type]
 
+    if source == "grand_calculated_summary":
```
MebinAbraham, line 63: A unit test needs adding to `test_types.py`.

katie-gardner: done

liamtoozer (approved):
Tested against the PR schema and works great for me 👍

berroar (approved):

app/validators/routing/types.py
berroar: Resolving on behalf of @MebinAbraham 👍

PR #230, opened 2023-12-11
Support float max or min

VirajP1002:

### How to review

### Checklist

### DIFF
--- a/app/validators/answers/number_answer_validator.py
+++ b/app/validators/answers/number_answer_validator.py
@@ -22,7 +22,7 @@ class NumberAnswerValidator(AnswerValidator):
         "The referenced answer has a greater number of decimal places than answer"
     )
     MAX_MIN_IS_STRING = (
-        "The minimum or maximum value is a string"
+        "The minimum or maximum value is not a float or an integer"
     )
 
     def __init__(self, schema_element, questionnaire_schema):
@@ -54,7 +54,6 @@ class NumberAnswerValidator(AnswerValidator):
             )
 
             self.validate_min_max_is_number()
-
             self.validate_value_in_limits()
 
         return self.errors
@@ -67,22 +66,12 @@ class NumberAnswerValidator(AnswerValidator):
         min_value = self.answer.get("minimum", {}).get("value", 0)
         max_value = self.answer.get("maximum", {}).get("value", 0)
 
-        # Get minimum / maximum and iterate through to ensure value is not a string
-        for x in [min_value, max_value]:
-            if isinstance(x, str):
-                self.add_error(
-                    self.MAX_MIN_IS_STRING
-                )
-#######################################################################################
-        # if isinstance(min_value, str):
-        #     self.add_error(
-        #         self.MINIMUM_IS_STRING
-        #     )
-        #
-        # if isinstance(max_value, str):
-        #     self.add_error(
-        #         self.MAXIMUM_IS_STRING
-        #     )
+        if min_value or max_value:  # Checks both min and max simultaneously
+            for value in [min_value, max_value]:
+                if not isinstance(value, float) or not isinstance(value, int):
+                    self.add_error(
+                        self.MAX_MIN_IS_STRING
+                    )
 
     def validate_value_in_limits(self):
         min_value = self.answer.get("minimum", {}).get("value", 0)
--- a/app/validators/answers/number_answer_validator.py
+++ b/app/validators/answers/number_answer_validator.py
@@ -21,6 +21,9 @@ class NumberAnswerValidator(AnswerValidator):
     GREATER_DECIMALS_ON_ANSWER_REFERENCE = (
         "The referenced answer has a greater number of decimal places than answer"
     )
+    MAX_MIN_IS_STRING = (
+        "The minimum or maximum value is a string"
+    )
 
     def __init__(self, schema_element, questionnaire_schema):
         super().__init__(schema_element, questionnaire_schema)
@@ -50,6 +53,8 @@ class NumberAnswerValidator(AnswerValidator):
                 self.questionnaire_schema.numeric_answer_ranges
             )
 
+            self.validate_min_max_is_number()
+
             self.validate_value_in_limits()
 
         return self.errors
@@ -57,6 +62,27 @@ class NumberAnswerValidator(AnswerValidator):
     def validate_mandatory_has_no_default(self):
         if self.answer.get("mandatory") and self.answer.get("default") is not None:
             self.add_error(self.DEFAULT_ON_MANDATORY)
+
+    def validate_min_max_is_number(self):
+        min_value = self.answer.get("minimum", {}).get("value", 0)
+        max_value = self.answer.get("maximum", {}).get("value", 0)
+
+        # Get minimum / maximum and iterate through to ensure value is not a string
+        for x in [min_value, max_value]:
+            if isinstance(x, str):
+                self.add_error(
+                    self.MAX_MIN_IS_STRING
+                )
+#######################################################################################
+        # if isinstance(min_value, str):
+        #     self.add_error(
+        #         self.MINIMUM_IS_STRING
+        #     )
+        #
+        # if isinstance(max_value, str):
+        #     self.add_error(
+        #         self.MAXIMUM_IS_STRING
+        #     )
 
     def validate_value_in_limits(self):
         min_value = self.answer.get("minimum", {}).get("value", 0)
--- a/app/validators/answers/number_answer_validator.py
+++ b/app/validators/answers/number_answer_validator.py
@@ -21,9 +21,7 @@ class NumberAnswerValidator(AnswerValidator):
     GREATER_DECIMALS_ON_ANSWER_REFERENCE = (
         "The referenced answer has a greater number of decimal places than answer"
     )
-    MAX_MIN_IS_STRING = (
-        "The minimum or maximum value is not a float or an integer"
-    )
+    MAX_MIN_IS_STRING = "The minimum or maximum value is not a float or an integer"
 
     def __init__(self, schema_element, questionnaire_schema):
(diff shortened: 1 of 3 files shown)


app/validators/answers/number_answer_validator.py
```diff
@@ -58,6 +65,15 @@ def validate_mandatory_has_no_default(self):
         if self.answer.get("mandatory") and self.answer.get("default") is not None:
             self.add_error(self.DEFAULT_ON_MANDATORY)
 
+    def validate_min_max_is_number(self):
+        min_value = self.answer.get("minimum", {}).get("value", 0)
+        max_value = self.answer.get("maximum", {}).get("value", 0)
```
katie-gardner, line 70: This isn't going to work for min/max that is referring to a value source, at the moment if a value source is 
``` 
value: {source: ..., identifier: ...}
```
the test passes because the type of value is dict, which is not a string, so this needs to be checking for a numeric type instead of not being a string, and may need to do something similar to runner for having a map of min & max

tests/validators/answers/test_number_answer_validator.py
```diff
@@ -80,6 +80,118 @@ def test_are_decimal_places_valid():
...
+                "source": "calculated_summary",
+                "identifier": "currency-total-playback",
+            }
+        },
+    }
+
+    validator = NumberAnswerValidator(
+        answer, get_mock_schema_with_data_version("0.0.3")
+    )
+
```
katie-gardner, line 191: the validator never runs which is why the errors are 0, it needs to run before errors are checked

tests/validators/answers/test_number_answer_validator.py
```diff
@@ -80,6 +80,118 @@ def test_are_decimal_places_valid():
...
+        "description": "This is a description of the maximum value",
+        "label": "Set a value less than the total above",
+        "mandatory": True,
+        "type": "Currency",
+        "currency": "GBP",
+        "decimal_places": 2,
+        "maximum": {
+            "value": {
+                "source": "calculated_summary",
+                "identifier": "currency-total-playback",
```
katie-gardner, line 183: This also only passes for the reason above where value is a dictionary, I think you'll need to use a real schema for this test so you can correctly determine that the value source is of a numeric type

katie-gardner (dismissed):
This will need something a little different to handle value sources that reference another answer, you'll likely need to look that up in one of the existing questionnaire schema maps to find the type of the source

tests/validators/answers/test_number_answer_validator.py
VirajP1002: Thanks for spotting, added that in

app/validators/answers/number_answer_validator.py
```diff
@@ -29,6 +35,12 @@ def __init__(self, schema_element, questionnaire_schema):
     def validate(self):
         super().validate()
 
+        self.validate_min_max_is_number()
+
+        # Prevent other validation methods that requires calculations running into errors due to types
+        if len(self.errors) > 0:
```
katie-gardner, line 41: You don't actually need the `len`, since `self.errors` is a list you can just do 
```
if self.errors:
    return self.errors
```
its truthy if the list contains items, and falsy otherwise

app/validators/answers/number_answer_validator.py
```diff
@@ -58,6 +70,20 @@ def validate_mandatory_has_no_default(self):
         if self.answer.get("mandatory") and self.answer.get("default") is not None:
             self.add_error(self.DEFAULT_ON_MANDATORY)
 
+    def validate_min_max_is_number(self):
+        for min_max in ["minimum", "maximum"]:
+            value = self.answer.get(min_max, {}).get("value", 0)
+            if value:
```
katie-gardner, line 76: You can compress these two lines into
`if value := self.answer.get(min_max, {}).get("value"):`
(no need to default to 0 in this case)

app/validators/routing/types.py
```diff
@@ -42,33 +42,67 @@
...
+
+
+def resolve_list_source_json_type(value_source):
+    if selector := value_source.get("selector"):
+        return LIST_SELECTOR_TO_JSON_TYPE[selector]
+    return TYPE_ARRAY
+
 
 def resolve_value_source_json_type(value_source, schema):
     source = value_source["source"]
```
katie-gardner, line 91: I would pull out both the source _and the identifier_ at this stage and then pass the identifier to the functions, saves you doing it in each one
(+ each of your new functions only needs the identifier, so nice to not pass in the whole value source)

app/validators/routing/types.py
```diff
@@ -42,33 +42,67 @@
...
+}
+
+
+def resolve_answer_source_json_type(value_source, schema):
+    answer_id = value_source["identifier"]
+    answer_type = schema.answers_with_context[answer_id]["answer"]["type"]
+    return ANSWER_TYPE_TO_JSON_TYPE[answer_type]
+
+
+def resolve_calculated_summary_source_json_type(value_source, schema):
```
katie-gardner, line 58: based off my comment on identifier/source the arguments for this could then be 
`resolve_calculated_summary_source_json_type(block_id, schema):`

tests/validators/answers/test_number_answer_validator.py
```diff
@@ -80,6 +80,107 @@ def test_are_decimal_places_valid():
     }
 
 
+def test_maximum_set_as_string():
+    answer = {
```
katie-gardner, line 84: for absolute clarity I might call this function
`test_invalid_maximum_value_is_string` and similar for the ones below

tests/validators/answers/test_number_answer_validator.py
```diff
@@ -80,6 +80,107 @@ def test_are_decimal_places_valid():
     }
 
 
+def test_maximum_set_as_string():
```
katie-gardner, line 83: Because these 3 tests are identical except for the minimum and maximum, you could parameterize them 
```
@pytest.mark.parameterize("minimum, maximum", [({"value": "100"}, None), (None, {"value": "0"}), ({"value": "100", {"value": "0"})])
```
and cut the repeated code down

tests/validators/answers/test_number_answer_validator.py
```diff
@@ -80,6 +80,107 @@ def test_are_decimal_places_valid():
...
+    validator.validate_min_max_is_number()
+
+    assert validator.errors[1] == {
+        "message": validator.MIN_OR_MAX_IS_NOT_NUMERIC,
+        "answer_id": "total-percentage",
+    }
+    assert len(validator.errors) == 2
+
+
+def test_min_if_set_as_float():
```
katie-gardner, line 152: As above, maybe `test_valid_minimum_value_is_float` for clarity, so the tests indicate if its a valid/invalid case

tests/validators/routing/test_types.py
```diff
@@ -98,6 +98,19 @@ def test_resolve_grand_calculated_summary_value_source_json_type():
     )
 
 
+def test_resolve_metadata_summary_value_source_json_type():
```
katie-gardner, line 101: I'd parametrize this test too, and do a case for each of the different possible metadata json types

katie-gardner (commented):
A few minor things following the fixes, but looking good!

katie-gardner (commented):
Since we both looked at it, I'll leave to 2 others as the official reviewers, but looking good to me! 🎉

app/validators/answers/number_answer_validator.py
```diff
@@ -58,6 +70,19 @@ def validate_mandatory_has_no_default(self):
...
+    def validate_min_max_is_number(self):
+        for min_max in ["minimum", "maximum"]:
+            if value := self.answer.get(min_max, {}).get("value", 0):
+                if isinstance(value, dict):
+                    if (
+                        resolve_value_source_json_type(value, self.questionnaire_schema)
+                        != TYPE_NUMBER
+                    ):
+                        self.add_error(self.MIN_OR_MAX_IS_NOT_NUMERIC)
+                else:
```
berroar, line 82: Could use elif here?

app/validators/routing/types.py
```diff
@@ -42,33 +42,67 @@
...
 }
 
+METADATA_TYPE_TO_JSON_TYPE = {
+    "string": TYPE_STRING,
+    "date": TYPE_STRING,
+    "boolean": TYPE_BOOLEAN,
+}
+
+
+def resolve_answer_source_json_type(answer_id, schema):
```
berroar, line 52: Probably worth adding type hints to methods like these while refactoring 👍

app/validators/routing/types.py
```diff
@@ -42,33 +42,67 @@
...
+        for values in schema.schema.get("metadata", []):
+            if values.get("name") == identifier:
+                return METADATA_TYPE_TO_JSON_TYPE[values.get("type")]
+    return TYPE_STRING
+
+
+def resolve_list_source_json_type(selector):
+    if selector:
+        return LIST_SELECTOR_TO_JSON_TYPE[selector]
+    return TYPE_ARRAY
```
berroar, line 85: This could be simplified
```suggestion
return LIST_SELECTOR_TO_JSON_TYPE[selector] if selector else TYPE_ARRAY
```

berroar (approved):

tests/schemas/rules/boolean/comparison/equal.json
```diff
@@ -34,5 +34,17 @@
...
+  },
+  {
+    "description": "Compare an answer to a boolean",
+    "rule": {
+      "==": [
+        {
+          "source": "answers",
+          "identifier": "test"
+        },
+        true
```
MebinAbraham, line 46: We don't have boolean answers?

MebinAbraham (approved):

Yuyuutsu (commented):
Changes look good 🎉

PR #243, opened 2024-06-04
Add list source reference order validation

petechd:
Adds validation of sources referencing lists, no longer list source for a given list can be referenced in a section that precedes the section in which the list collector itself for that list is set.

### How to review
New invalid schema added and new unit test that checks the error message.

### Checklist

### DIFF
--- a/app/error_messages.py
+++ b/app/error_messages.py
@@ -40,6 +40,4 @@ VARIANTS_HAVE_MULTIPLE_QUESTION_TYPES = (
     "Variants have more than one question type for block."
 )
 PREVIEW_WITHOUT_INTRODUCTION_BLOCK = "No introduction block found. Introduction block is mandatory when using the preview questions feature."
-LIST_REFERENCED_BEFORE_ADDED = (
-    "List '{list_name}' referenced as source before it has been created."
-)
+LIST_REFERENCED_BEFORE_ADDED = "List referenced as source before it has been created."
--- a/app/error_messages.py
+++ b/app/error_messages.py
@@ -40,3 +40,6 @@ VARIANTS_HAVE_MULTIPLE_QUESTION_TYPES = (
     "Variants have more than one question type for block."
 )
 PREVIEW_WITHOUT_INTRODUCTION_BLOCK = "No introduction block found. Introduction block is mandatory when using the preview questions feature."
+LIST_REFERENCED_BEFORE_ADDED = (
+    "List '{list_name}' referenced as source before it has been created."
+)
--- a/app/error_messages.py
+++ b/app/error_messages.py
@@ -40,4 +40,4 @@ VARIANTS_HAVE_MULTIPLE_QUESTION_TYPES = (
     "Variants have more than one question type for block."
 )
 PREVIEW_WITHOUT_INTRODUCTION_BLOCK = "No introduction block found. Introduction block is mandatory when using the preview questions feature."
-LIST_REFERENCED_BEFORE_ADDED = "List referenced as source before it has been created."
+LIST_REFERENCED_BEFORE_CREATED = "List referenced as source before it has been created."
--- a/app/validators/questionnaire_schema.py
+++ b/app/validators/questionnaire_schema.py
@@ -227,7 +227,7 @@ class QuestionnaireSchema:
                 list_id = block["for_list"]
                 if list_id not in self._lists_with_context:
                     section_id = self.get_section_id_for_block_id(block["id"])
-                    section_index = self.get_section_index_for_section_id(section_id)
+                    section_index = self.section_ids.index(section_id)
                     self._lists_with_context[list_id] = {
                         "section_index": section_index,
                         "block_index": self.block_ids.index(block["id"]),
@@ -240,7 +240,7 @@ class QuestionnaireSchema:
                     < self._lists_with_context[list_id]["block_index"]
                 ):
                     section_id = self.get_section_id_for_block_id(block["id"])
-                    section_index = self.get_section_index_for_section_id(section_id)
+                    section_index = self.section_ids.index(section_id)
                     self._lists_with_context[list_id] = {
                         "section_index": section_index,
                         "block_index": self.block_ids.index(block["id"]),
@@ -656,26 +656,3 @@ class QuestionnaireSchema:
     def get_section_id_for_block_id(self, block_id: str) -> str | None:
         if block := self.get_block(block_id):
             return self.get_section_id_for_block(block)
-
-    def get_list_collector_block_id_for_add_block(self, block_id) -> dict | None:
-        for block in self.blocks:
-            if (
-                block["type"] == "ListCollector"
-                and block["add_block"]["id"] == block_id
-            ):
-                return block["id"]
-
-    def get_list_collector_block_id_for_repeating_blocks(self, block_id) -> dict | None:
-        for block in self.blocks:
-            if block["type"] in [
-                "ListCollector",
-                "ListCollectorContent",
-            ] and block.get("repeating_blocks"):
-                for repeating_block in block["repeating_blocks"]:
-                    if repeating_block["id"] == block_id:
-                        return block["id"]
-
-    def get_section_index_for_section_id(self, section_id: str) -> int:
-        for index, section in enumerate(self.sections):
-            if section["id"] == section_id:
-                return index
--- a/app/validators/questionnaire_schema.py
+++ b/app/validators/questionnaire_schema.py
@@ -619,24 +619,22 @@ class QuestionnaireSchema:
             return self.get_section_id_for_block(block)
 
     def get_list_collector_block_id_for_add_block(self, block_id) -> dict | None:
-        for section_id, blocks in self.blocks_by_section_id.items():
-            for block in blocks:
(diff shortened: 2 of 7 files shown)


tests/test_questionnaire_validator.py
```diff
@@ -406,3 +406,21 @@ def test_invalid_calculated_or_grand_calculated_summary_id_in_value_source():
     validator.validate()
 
     assert validator.errors == expected_errors
+
+
+def test_list_as_source_referenced_before_created():
```
berroar, line 411: As discussed, will need to cover the case when the list source is before the collector within the same section also

petechd: Thanks, all seems to be added, there might be some very edge cases here and there but I think all is checked now.

app/validators/questionnaire_validator.py
```diff
@@ -134,3 +139,90 @@ def validate_introduction_block(self):
...
+                        # Section level "enabled" rule that can use list source,
+                        # check: common_definitions.json#/section_enabled
+                        self.add_error(
+                            error_messages.LIST_REFERENCED_BEFORE_ADDED.format(
+                                list_name=list_identifier
+                            ),
+                            section_name=section["id"],
+                        )
+
+    def populate_referenced_lists(self):
```
berroar, line 180: Since this is just populating, I would move this to questionnaire_schema and populate the map up front similar to what we the way we do in runner

petechd: I made it questionnaire schema `@property` similar to `answers_with_context`. Thanks.

app/error_messages.py
```diff
@@ -40,3 +40,6 @@
     "Variants have more than one question type for block."
 )
 PREVIEW_WITHOUT_INTRODUCTION_BLOCK = "No introduction block found. Introduction block is mandatory when using the preview questions feature."
+LIST_REFERENCED_BEFORE_ADDED = (
+    "List '{list_name}' referenced as source before it has been created."
```
liamtoozer, line 44: Quick question - could the list ID be passed in as a key/val to the error message instead, to prevent using string formatting?

app/validators/questionnaire_schema.py
```diff
@@ -209,6 +210,44 @@ def answers_with_context(self):
...
+                    "section_index": 0,
+                    "block_index": 0,
+                }
+
+        if blocks := self.list_collectors:
+            for block in blocks:
+                list_id = block["for_list"]
+                if list_id not in self._lists_with_context:
+                    section_id = self.get_section_id_for_block_id(block["id"])
+                    section_index = self.get_section_index_for_section_id(section_id)
```
liamtoozer, line 230: Minor learning question! Can `section_index = self.section_ids.index(section_id)` be used here so that a function might not be needed? Or is that not a good idea as it might depend on the order?

app/validators/questionnaire_schema.py
```diff
@@ -617,3 +656,26 @@ def get_section_id_for_block(self, block: Mapping) -> str | None:
     def get_section_id_for_block_id(self, block_id: str) -> str | None:
         if block := self.get_block(block_id):
             return self.get_section_id_for_block(block)
+
+    def get_list_collector_block_id_for_add_block(self, block_id) -> dict | None:
```
liamtoozer, line 660: I couldn't spot this being used, so removing it might be okay?

app/validators/questionnaire_schema.py
```diff
@@ -617,3 +656,26 @@ def get_section_id_for_block(self, block: Mapping) -> str | None:
...
+        for block in self.blocks:
+            if (
+                block["type"] == "ListCollector"
+                and block["add_block"]["id"] == block_id
+            ):
+                return block["id"]
+
+    def get_list_collector_block_id_for_repeating_blocks(self, block_id) -> dict | None:
```
liamtoozer, line 668: I couldn't spot this being used, so removing it might be okay?

app/validators/questionnaire_validator.py
```diff
@@ -134,3 +139,38 @@ def validate_introduction_block(self):
...
+                    if parent_block:
+                        if (
+                            self.questionnaire_schema.block_ids.index(
+                                parent_block["id"]
+                            )
+                            < lists_with_context[list_identifier]["block_index"]
+                        ):
+                            self.add_error(
+                                error_messages.LIST_REFERENCED_BEFORE_ADDED.format(
+                                    list_name=list_identifier
```
liamtoozer, line 161: Might not be possible, but could the parent block ID also be output in this error message, just for some extra info to help narrow down the location in the schema?

tests/test_questionnaire_validator.py
```diff
@@ -406,3 +406,74 @@ def test_invalid_calculated_or_grand_calculated_summary_id_in_value_source():
...
+    filename = "schemas/invalid/test_invalid_list_source_reference.json"
+    validator = QuestionnaireValidator(_open_and_load_schema_file(filename))
+
+    expected_errors = [
+        {
+            "message": "List 'household' referenced as source before it has been "
```
liamtoozer, line 417: V minor/cosmetic - each `message` on a single line?

liamtoozer (commented):
Looks great to me, and thanks for guiding me through the changes! Just had some extra questions

app/validators/questionnaire_schema.py
liamtoozer: Ah, unless this is for the other PR coming up?

app/validators/questionnaire_schema.py
liamtoozer: Ah, unless this is for the other PR coming up?

app/error_messages.py
petechd: Yeah, we can do it both ways, up to the devs which way we want this.

app/validators/questionnaire_schema.py
petechd: No, it's not needed after these simplifications I showed you on Monday, not even needed for that next PR coming up.

app/validators/questionnaire_schema.py
petechd: Again, it's not needed after these simplifications I showed you on Monday, not even needed for that next PR coming up.

app/validators/questionnaire_schema.py
petechd: This one is a really good one, how did I miss that opportunity there? Will use your suggestion in both places.

app/validators/questionnaire_validator.py
petechd: Not a problem,  I will add it as a key/value.

petechd: We could possibly keep it as "parent block id" since the block id could be sometimes the nested one, like the add block one.

petechd: I've added it as "parent block id" for now.

tests/test_questionnaire_validator.py
petechd: They are now are after we removed list name to it's own key/value pair, thanks.

app/error_messages.py
petechd: I moved it to a key/val and lets see if other devs like it.

tests/test_questionnaire_validator.py
petechd: We can pass the actual error message constant here now, import it, but only when they list id key/val is separated out so not sure if we want to separate it out? @berroar?

berroar: I  think splitting it out is fine (e.g. the current one) because at the end of the day, it is a case of clearly presenting the info to an author, so looks good to me 👍

petechd: Ok, added the changes now, thanks.

app/validators/questionnaire_schema.py
```diff
@@ -209,6 +210,44 @@ def answers_with_context(self):
         self._answers_with_context = answers_dict
         return self._answers_with_context
 
+    @property
+    def lists_with_context(self):
+        if self._lists_with_context:
+            return self._lists_with_context
```
berroar, line 216: Do we need this if/return, I don't think we would ever hit this at the moment?

app/error_messages.py
```diff
@@ -40,3 +40,4 @@
     "Variants have more than one question type for block."
 )
 PREVIEW_WITHOUT_INTRODUCTION_BLOCK = "No introduction block found. Introduction block is mandatory when using the preview questions feature."
+LIST_REFERENCED_BEFORE_ADDED = "List referenced as source before it has been created."
```
berroar, line 43: Really minor but would make it consistent and use created rather than added as the var name too.

app/validators/questionnaire_schema.py
petechd: I went with the same approach we use in answers_with_context, since this is a property, we populate it only the initial time.

petechd: I've added the change now after our discussion.

Yuyuutsu (approved):
Validation works as expected, LGTM !

berroar (approved):

liamtoozer (approved):

PR #244, opened 2024-06-04
Add validation for invalid html tags

berroar:

### How to review

### Checklist

### DIFF
--- a/app/validators/questionnaire_validator.py
+++ b/app/validators/questionnaire_validator.py
@@ -30,6 +30,7 @@ class QuestionnaireValidator(Validator):
         self.validate_duplicates()
         self.validate_smart_quotes()
         self.validate_white_spaces()
+        self.validate_html()
 
         for section in self.questionnaire_schema.sections:
             section_validator = SectionValidator(section, self.questionnaire_schema)
@@ -106,6 +107,43 @@ class QuestionnaireValidator(Validator):
                         error_messages.DUMB_QUOTES_FOUND,
                         pointer=translatable_item.pointer,
                     )
+
+    def validate_html(self):
+        html_strings = []
+        schema_object = SurveySchema(self.schema_element)
+
+        # pylint: disable=invalid-string-quote
+        html_regex = re.compile(r"<[^>]*>")
+
+        for translatable_item in schema_object.translatable_items:
+            schema_text = translatable_item.value
+
+            values_to_check = [schema_text]
+
+            if isinstance(schema_text, dict):
+                values_to_check = schema_text.values()
+
+            for schema_text in values_to_check:
+                if schema_text and html_regex.search(schema_text):
+                    html_strings.append(
+                        {"pointer": translatable_item.pointer, "text": schema_text}
+                    )
+
+        if html_strings:
+            self.check_invalid_html_tags(html_strings)
+
+    def check_invalid_html_tags(self, html_strings):
+        strong = re.compile("<strong[^>]>*>|</strong>")
+        anchor = re.compile("<a[^>]>*>|</a>")
+        for html_string in html_strings:
+            if not strong.search(html_string["text"]) and not anchor.search(
+                html_string["text"]
+            ):
+                self.add_error(
+                    error_messages.HTML_FOUND,
+                    pointer=html_string["pointer"],
+                    text=html_string["text"],
+                )
 
     def validate_white_spaces(self):
         schema_object = SurveySchema(self.schema_element)
--- a/tests/schemas/invalid/test_invalid_html_in_schema_text.json
+++ b/tests/schemas/invalid/test_invalid_html_in_schema_text.json
@@ -5,7 +5,7 @@
   "data_version": "0.0.3",
   "survey_id": "144",
   "theme": "default",
-  "title": "Test introduction",
+  "title": "Test invalid html",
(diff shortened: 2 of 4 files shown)


VirajP1002 (approved):
Works really nicely! Tests pass and changes look good. I also tested the changes in Runner and validation works as intended!

app/validators/questionnaire_validator.py
```diff
@@ -107,6 +108,43 @@ def validate_smart_quotes(self):
...
+                if schema_text and html_regex.search(schema_text):
+                    html_strings.append(
+                        {"pointer": translatable_item.pointer, "text": schema_text}
+                    )
+
+        if html_strings:
+            self.check_invalid_html_tags(html_strings)
+
+    def check_invalid_html_tags(self, html_strings):
+        strong = re.compile("<strong[^>]>*>|</strong>")
```
petechd, line 136: I don't think `strong` needs the same regex `a` has as `strong` doesnt accept any extra attributes, it's always just plain `strong` tags.

petechd: Maybe something like: `<strong>.*<\/strong>`?

app/validators/questionnaire_validator.py
```diff
@@ -107,6 +108,43 @@ def validate_smart_quotes(self):
...
+        anchor = re.compile("<a[^>]>*>|</a>")
```
petechd, line 137: Possibly can make it more precise by using something like: `<a .*>.*<\/a>`?

app/validators/questionnaire_validator.py
```diff
@@ -107,6 +108,43 @@ def validate_smart_quotes(self):
...
+        for html_string in html_strings:
+            if not strong.search(html_string["text"]) and not anchor.search(
```
petechd, line 139: At the moment some other html tags can pass this validation if they are in the same string as `a` and `strong` so maybe we should check how many of these we have combined compared to total amount of html tags present and raise an error if they are not the same?
```python
    all_tags = re.compile("<([a-z0-9]+)(?=[\s>])(?:[^>=]|='[^']*'|=\"[^\"]*\"|=[^'\"\s]*)*\s?\/?>")

    if len(strong.findall(html)) + len(anchor.findall(html)) != len(all_tags.findall(html)):
```

app/validators/questionnaire_validator.py
berroar: Updated this but had to make a slight amendment for when we have multiple strongs in one string as discussed 👍

app/validators/questionnaire_validator.py
berroar: Updated and added some additional cases and tests with the schema 👍

app/validators/questionnaire_validator.py
berroar: Also handling potential multiple instances now as per strong change

app/validators/questionnaire_validator.py
```diff
@@ -132,10 +132,10 @@ def validate_html(self):
             self.check_invalid_html_tags(html_strings)
 
     def check_invalid_html_tags(self, html_strings):
-        strong = re.compile(r"<strong>(?:(?!<\/strong>).)*</strong>")
-        anchor = re.compile("<a .*>.*</a>")
+        strong = re.compile(r"<strong>(?:(?!</strong>).)*</strong>")
+        anchor = re.compile("<a .*>(?:(?!</a).)*</a>")
```
petechd, line 136: Raw string missing from the anchor expression.

berroar: Updated 👍

tests/schemas/invalid/test_invalid_html_in_schema_text.json
```diff
@@ -0,0 +1,121 @@
...
+                    "description": "<h1>Title</h1><em>Not valid tag</em>"
+                  }
+                ]
+              }
+            },
+            {
+              "type": "Interstitial",
+              "id": "interstitial-four",
+              "content": {
+                "title": "Valid double <strong>strong</strong> with another <strong>strong</strong>.",
```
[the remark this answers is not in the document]
berroar: Added but didn't involve a minor fix 😅

app/validators/questionnaire_validator.py
```diff
@@ -107,6 +108,47 @@ def validate_smart_quotes(self):
...
+                {"pointer": translatable_item.pointer, "text": schema_text}
+                for schema_text in values_to_check
+                if schema_text and html_regex.search(schema_text)
+            )
+        if html_strings:
+            self.check_invalid_html_tags(html_strings)
+
+    def check_invalid_html_tags(self, html_strings):
+        strong = re.compile(r"<strong>(?:(?!</strong>).)*</strong>")
+        anchor = re.compile(r"<a [^>]*>.*?</a>")
```
petechd, line 136: Since anchor [always takes `href` argument only](https://www.w3schools.com/tags/att_a_href.asp) can we add the href as something always present in this regex?:
```suggestion
        anchor = re.compile(r"<a href[^>]*>.*?</a>")
```

petechd (approved):

app/validators/questionnaire_validator.py
berroar: Reverted as it leads to some false positives where href is present but not the first argument

petechd (approved):

PR #245, opened 2024-06-11
Add answer source reference order validation

petechd:
Adds validation of objects referencing answers sources, no longer answer source for a given block can be referenced in a section that precedes the section in which the answer itself was set. Order of Block indexes is also checked, as well as group and section level answer source rules like "enabled" and "skip conditions".

### How to review
New invalid schema added and new unit test that checks the error message.

### Checklist

### DIFF
--- a/app/error_messages.py
+++ b/app/error_messages.py
@@ -40,6 +40,6 @@ VARIANTS_HAVE_MULTIPLE_QUESTION_TYPES = (
     "Variants have more than one question type for block."
 )
 PREVIEW_WITHOUT_INTRODUCTION_BLOCK = "No introduction block found. Introduction block is mandatory when using the preview questions feature."
-ANSWER_REFERENCED_BEFORE_ADDED = (
+ANSWER_REFERENCED_BEFORE_EXISTS = (
     "Answer '{answer_id}' referenced as source before it has been added."
 )
--- a/app/error_messages.py
+++ b/app/error_messages.py
@@ -40,3 +40,6 @@ VARIANTS_HAVE_MULTIPLE_QUESTION_TYPES = (
     "Variants have more than one question type for block."
 )
 PREVIEW_WITHOUT_INTRODUCTION_BLOCK = "No introduction block found. Introduction block is mandatory when using the preview questions feature."
+ANSWER_REFERENCED_BEFORE_ADDED = (
+    "Answer '{answer_id}' referenced as source before it has been added."
+)
--- a/app/validators/questionnaire_schema.py
+++ b/app/validators/questionnaire_schema.py
@@ -621,7 +621,7 @@ class QuestionnaireSchema:
                     return self.sections_by_id[section_id]
 
     def get_parent_list_collector_for_add_block(self, block_id) -> dict | None:
-        for _, blocks in self.blocks_by_section_id.items():
+        for blocks in self.blocks_by_section_id.values():
             for block in blocks:
                 if (
                     block["type"] == "ListCollector"
@@ -630,7 +630,7 @@ class QuestionnaireSchema:
                     return block["id"]
 
     def get_parent_list_collector_for_repeating_block(self, block_id) -> dict | None:
-        for _, blocks in self.blocks_by_section_id.items():
+        for blocks in self.blocks_by_section_id.values():
             for block in blocks:
                 if block["type"] in [
                     "ListCollector",
@@ -639,6 +639,7 @@ class QuestionnaireSchema:
                     for repeating_block in block["repeating_blocks"]:
                         if repeating_block["id"] == block_id:
                             return block["id"]
+        return None
 
     def is_block_in_repeating_section(self, block_id: str) -> bool:
         parent_section = self.get_parent_section_for_block(block_id)
--- a/app/validators/questionnaire_schema.py
+++ b/app/validators/questionnaire_schema.py
@@ -585,7 +585,7 @@ class QuestionnaireSchema:
                     return self.sections_by_id[section_id]
 
     def get_parent_list_collector_for_add_block(self, block_id) -> dict | None:
-        for section_id, blocks in self.blocks_by_section_id.items():
+        for _, blocks in self.blocks_by_section_id.items():
             for block in blocks:
                 if (
                     block["type"] == "ListCollector"
@@ -594,7 +594,7 @@ class QuestionnaireSchema:
                     return block["id"]
 
     def get_parent_list_collector_for_repeating_block(self, block_id) -> dict | None:
-        for section_id, blocks in self.blocks_by_section_id.items():
+        for _, blocks in self.blocks_by_section_id.items():
             for block in blocks:
                 if block["type"] in [
                     "ListCollector",
--- a/app/validators/questionnaire_schema.py
+++ b/app/validators/questionnaire_schema.py
@@ -587,13 +587,19 @@ class QuestionnaireSchema:
     def get_parent_list_collector_for_add_block(self, block_id) -> dict | None:
         for section_id, blocks in self.blocks_by_section_id.items():
             for block in blocks:
-                if block["type"] == "ListCollector" and block["add_block"]["id"] == block_id:
+                if (
+                    block["type"] == "ListCollector"
+                    and block["add_block"]["id"] == block_id
+                ):
                     return block["id"]
 
     def get_parent_list_collector_for_repeating_block(self, block_id) -> dict | None:
         for section_id, blocks in self.blocks_by_section_id.items():
             for block in blocks:
-                if block["type"] in ["ListCollector", "ListCollectorContent"] and block.get("repeating_blocks"):
+                if block["type"] in [
+                    "ListCollector",
+                    "ListCollectorContent",
+                ] and block.get("repeating_blocks"):
                     for repeating_block in block["repeating_blocks"]:
                         if repeating_block["id"] == block_id:
                             return block["id"]
--- a/app/validators/questionnaire_schema.py
+++ b/app/validators/questionnaire_schema.py
@@ -584,6 +584,20 @@ class QuestionnaireSchema:
                 if block_id == block["id"]:
                     return self.sections_by_id[section_id]
 
+    def get_parent_list_collector_for_add_block(self, block_id) -> dict | None:
+        for section_id, blocks in self.blocks_by_section_id.items():
+            for block in blocks:
+                if block["type"] == "ListCollector" and block["add_block"]["id"] == block_id:
+                    return block["id"]
+
+    def get_parent_list_collector_for_repeating_block(self, block_id) -> dict | None:
+        for section_id, blocks in self.blocks_by_section_id.items():
+            for block in blocks:
+                if block["type"] in ["ListCollector", "ListCollectorContent"] and block.get("repeating_blocks"):
+                    for repeating_block in block["repeating_blocks"]:
+                        if repeating_block["id"] == block_id:
+                            return block["id"]
+
     def is_block_in_repeating_section(self, block_id: str) -> bool:
         parent_section = self.get_parent_section_for_block(block_id)
         return parent_section and self.is_repeating_section(parent_section["id"])
@@ -617,3 +631,8 @@ class QuestionnaireSchema:
     def get_section_id_for_block_id(self, block_id: str) -> str | None:
         if block := self.get_block(block_id):
             return self.get_section_id_for_block(block)
+
+    def get_section_index_for_section_id(self, section_id: str) -> int:
+        for index, section in enumerate(self.sections):
+            if section["id"] == section_id:
+                return index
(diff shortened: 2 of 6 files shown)


tests/schemas/invalid/test_invalid_answer_source_reference.json
```diff
@@ -0,0 +1,239 @@
...
+      }
+    }
+  },
+  "sections": [
+    {
+      "id": "confirmation-section",
+      "title": "Questions",
+      "groups": [
+        {
+          "id": "confirmation-block",
```
berroar, line 46: How come we output the errors at the group level rather than the block with the reference? Also confusing that the group is called `confirmation-block`, particularly in the tests

app/error_messages.py
```diff
@@ -40,3 +40,6 @@
     "Variants have more than one question type for block."
 )
 PREVIEW_WITHOUT_INTRODUCTION_BLOCK = "No introduction block found. Introduction block is mandatory when using the preview questions feature."
+ANSWER_REFERENCED_BEFORE_ADDED = (
```
berroar, line 43: Not sure about `before_added` here again, as per the list PR, maybe it needs to before it exists or before answered 🤔

tests/test_questionnaire_validator.py
```diff
@@ -406,3 +410,40 @@ def test_invalid_calculated_or_grand_calculated_summary_id_in_value_source():
     validator.validate()
 
     assert validator.errors == expected_errors
+
+
+def test_answer_as_source_referenced_before_created():
+    filename = "schemas/invalid/test_invalid_answer_source_reference.json"
```
berroar, line 416: Have tested and it works, but the test/schema needs to include an example within a repeating section 👍

tests/schemas/invalid/test_invalid_answer_source_reference.json
petechd: That's because  I've checked and the group can have skip conditions (as usual, can be based on answers) so I wanted that to be included in a schema as a theoretical example.

app/error_messages.py
petechd: This one is hard to get right for me.. "..BEFORE_EXISTS"?

tests/schemas/invalid/test_invalid_answer_source_reference.json
petechd: It's something I overlooked, now they are properly displaying block, group or section id depending what level object we validate.

berroar: I think the change has fixed this now

tests/test_questionnaire_validator.py
petechd: I've added repeating section, the only thing is it outputs multiple errors as there are question variants using answer source here.

tests/schemas/invalid/test_invalid_answer_source_reference.json
petechd: it outputs section, group or block depending on where the validated object is.

petechd:
I've added repeating blocks now to the test schema as well.

app/validators/questionnaire_validator.py
```diff
@@ -134,3 +140,115 @@ def validate_introduction_block(self):
...
+                source_block = self.questionnaire_schema.get_block_by_answer_id(
+                    identifier_reference["identifier"]
+                )
+                # Handling non-existing blocks used as source
+                if not source_block:
+                    self.add_error(
+                        ValueSourceValidator.ANSWER_SOURCE_REFERENCE_INVALID,
+                        identifier=identifier_reference["identifier"],
+                    )
+                    return False
```
liamtoozer, line 173: I had a learning question as I'm trying to understand the eQ schema layouts - Is there a reason why it doesn't continue the loop here, and instead returns on the first encounter of `ANSWER_SOURCE_REFERENCE_INVALID`?

app/validators/questionnaire_validator.py
```diff
@@ -134,3 +140,115 @@ def validate_introduction_block(self):
...
+                        )
+                    else:
+                        self.add_error(
+                            error_messages.ANSWER_REFERENCED_BEFORE_EXISTS.format(
+                                answer_id=identifier_reference["identifier"]
+                            ),
+                            group_d=group["id"],
+                        )
+
+    def validate_answer_source_section(self, section, section_index):
```
liamtoozer, line 208: Another learning question! Is it worth having type annotations for this function, or is that overkill?

app/validators/questionnaire_validator.py
petechd: Yeah, no worries, the `return` finishes the whole validation execution with a negative result (for a given schema), the `add_error` method only adds errors to the list. I copied it from another place, where a `True` or `False` output is expected, but I guess here we can continue the validation so I can remove the return.

app/validators/questionnaire_validator.py
petechd: If there aren't any in this module I wouldn't widen the scope of it by adding a task of refactoring it. Unless we want part of the module typed and part untyped. Up to the devs really..

app/validators/questionnaire_validator.py
liamtoozer: Cool that makes sense, thanks

app/validators/questionnaire_validator.py
liamtoozer: Yeah that makes sense, thanks

app/validators/questionnaire_validator.py
petechd: In this case we need to exit as the source block won't exist when later we try to access (access `source_block["type"]` it in:
```python
def resolve_source_block_id(self, source_block):
        # Handling of source block nested (list collector's add-block)
       if source_block["type"] == "ListAddQuestion":
```
so we need to exit early

liamtoozer: Ahh I see, okay cool, thanks!

app/validators/questionnaire_validator.py
```diff
@@ -140,6 +142,118 @@ def validate_introduction_block(self):
...
+                                answer_id=identifier_reference["identifier"]
+                            ),
+                            block_id=parent_block_id,
+                        )
+                    else:
+                        self.add_error(
+                            error_messages.ANSWER_REFERENCED_BEFORE_EXISTS.format(
+                                answer_id=identifier_reference["identifier"]
+                            ),
+                            group_d=group["id"],
```
berroar, line 206: Minor typo here

tests/test_questionnaire_validator.py
```diff
@@ -408,6 +412,66 @@ def test_invalid_calculated_or_grand_calculated_summary_id_in_value_source():
...
+        {
+            "block_id": "confirm-dob",
+            "message": "Answer 'date-of-birth-answer' referenced as source before it has "
+            "been added.",
+        },
+        {
+            "block_id": "confirm-dob",
+            "message": "Answer 'date-of-birth-answer' referenced as source before it has "
+            "been added.",
+        },
```
berroar, line 434: Only a minor comment, and up to you wether you want to change this, but from the surface these look like duplicates, but it is only when you check the test schema you can see that these are valid. Potentially could have added a couple of other blocks to be referenced as answers so that the id's are different, but only a minor thing.

app/validators/questionnaire_validator.py
petechd: Fixed!

tests/test_questionnaire_validator.py
Yuyuutsu: I have updated the schema to add a new `confirm-sex` block and removed some of the placeholders  `date-of-birth-answer `in the `confirm-dob` block to reduce the duplication.

tests/schemas/invalid/test_invalid_answer_source_reference.json
```diff
@@ -0,0 +1,1487 @@
...
+                        },
+                        {
+                          "placeholder": "sex-answer",
+                          "transforms": [
+                            {
+                              "arguments": {
+                                "source": "answers",
+                                "identifier": "sex-answer"
+                              },
+                              "transform": "copy"
```
berroar, line 683: I don't think this is a valid transform? Also I'm not sure why the new question we added is also a variant, as the whole point was to remove the confusing duplication from the test.

tests/schemas/invalid/test_invalid_answer_source_reference.json
```diff
@@ -0,0 +1,1487 @@
...
+                      "placeholders": [
+                        {
+                          "placeholder": "sex-answer",
+                          "transforms": [
+                            {
+                              "arguments": {
+                                "source": "answers",
+                                "identifier": "sex-answer"
+                              },
+                              "transform": "copy"
```
berroar, line 605: Here also 😅, not sure why this has been added?

app/validators/questionnaire_validator.py
```diff
@@ -140,6 +142,118 @@ def validate_introduction_block(self):
...
+                        )
+                    else:
+                        self.add_error(
+                            error_messages.ANSWER_REFERENCED_BEFORE_EXISTS.format(
+                                answer_id=identifier_reference["identifier"]
+                            ),
+                            group_id=group["id"],
+                        )
+
+    def validate_answer_source_section(self, section, section_index):
```
liamtoozer, line 209: I might be looking in the wrong place but would this need a new test? Or is it okay for coverage?

liamtoozer (commented):
Looks good to me! Just had some very minor comments/questions

app/validators/questionnaire_validator.py
```diff
@@ -140,6 +142,118 @@ def validate_introduction_block(self):
...
+
+        # Handling blocks in group
+        for group in self.questionnaire_schema.groups:
+            self.validate_answer_source_group(group)
+
+        # Handling section level "enabled" rule
+        for index, section in enumerate(self.questionnaire_schema.sections):
+            self.validate_answer_source_section(section, index)
+
+    def validate_answer_source_group(self, group):
```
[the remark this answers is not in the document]
petechd: I think it is already tested in the invalid schema? Since these are checked for errors in a group in a schema?

app/validators/questionnaire_schema.py
```diff
@@ -620,6 +620,26 @@ def get_parent_section_for_block(self, block_id) -> dict | None:
...
+        for _, blocks in self.blocks_by_section_id.items():
+            for block in blocks:
+                if (
+                    block["type"] == "ListCollector"
+                    and block["add_block"]["id"] == block_id
+                ):
+                    return block["id"]
+
+    def get_parent_list_collector_for_repeating_block(self, block_id) -> dict | None:
+        for _, blocks in self.blocks_by_section_id.items():
```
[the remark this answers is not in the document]
petechd: Ok, I just like that `_` syntax/convention, forgot about the more convenient `.values()`.

app/validators/questionnaire_schema.py
```diff
@@ -620,6 +620,26 @@ def get_parent_section_for_block(self, block_id) -> dict | None:
...
+    def get_parent_list_collector_for_add_block(self, block_id) -> dict | None:
+        for _, blocks in self.blocks_by_section_id.items():
+            for block in blocks:
+                if (
+                    block["type"] == "ListCollector"
+                    and block["add_block"]["id"] == block_id
+                ):
+                    return block["id"]
+
+    def get_parent_list_collector_for_repeating_block(self, block_id) -> dict | None:
```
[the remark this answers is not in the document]
petechd: I will add `return None` at the end for clarity.

app/validators/questionnaire_schema.py
```diff
@@ -620,6 +620,26 @@ def get_parent_section_for_block(self, block_id) -> dict | None:
                 if block_id == block["id"]:
                     return self.sections_by_id[section_id]
 
+    def get_parent_list_collector_for_add_block(self, block_id) -> dict | None:
+        for _, blocks in self.blocks_by_section_id.items():
```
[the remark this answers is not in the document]
petechd: Yeah that is better, check https://github.com/ONSdigital/eq-questionnaire-validator/pull/245#discussion_r1677466314

app/validators/questionnaire_validator.py
petechd: Sort of same as: https://github.com/ONSdigital/eq-questionnaire-validator/pull/245#discussion_r1677464858

tests/schemas/invalid/test_invalid_answer_source_reference.json
petechd: I've removed transform altogether as I couldn't find any suitable for this situation so only "bare" placeholder left here.

tests/schemas/invalid/test_invalid_answer_source_reference.json
petechd: Removed transform but also all the question variants so only one example was left here in the end.

liamtoozer (approved):

PR #252, opened 2024-09-09
Refactor to use FastAPI

petechd:
To improve validator performance we want to use a framework that supports concurrency. Mainly changes to Flask libraries specific code were made, they were replaced with FastAPI equivalents. I've reduced the number of modules and put the endpoints logic in the main file, since the codebase for the initial app is minimal in size. Some new logic added since FastAPI modules behave slightly different, mainly around data received in requests sent.

### How to review
Run it with Runner locally, ideally by pulling validator docker images and using make commands like `make run-validator` etc. using docker images and relevant branch of Runner.

### Checklist


liamtoozer (commented):
I think the `run_app.sh` might need to be updated with the new command as well?

Yuyuutsu (commented):
Works as expected through Runner. 
This part of the Make Run Command does not work : `poetry run ./scripts/run_app.sh`

Dockerfile
```diff
@@ -17,7 +17,4 @@ RUN poetry install --no-dev
 
 EXPOSE 5000
 
-ENV FLASK_APP=api.py
-
-ENTRYPOINT flask run --host 0.0.0.0
-
+CMD ["gunicorn", "api:app", "-b", "0.0.0.0:5000", "--workers", "20", "--worker-class", "uvicorn.workers.UvicornWorker", "--timeout", "0"]
```
liamtoozer, line 20: 🎨 Very minor/cosmetic - Could maybe add line breaks for each 4 flags & values to help readability?

api.py
```diff
@@ -1,11 +1,97 @@
-from flask import Flask
```
liamtoozer, line 1: Appreciate this wasn't there before, but would it be worth adding type hints now that it's being rejigged a bit? Might be one for a future ticket though so no worries if not

api.py
```diff
@@ -1,11 +1,97 @@
...
+                ),
+            )
+
+
+async def validate_schema(data):
+    json_to_validate = None
+    if data:
+        if isinstance(data, str):
+            try:
+                json_to_validate = json.loads(data)
```
liamtoozer, line 56: Minor suggestion - Could do this with a ternary to avoid the `elif` at the end? Something like 
```
try:
    json_to_validate = json.loads(data) if isinstance(data, str) else data
    ...

```

api.py
```diff
@@ -1,11 +1,97 @@
...
+            return response, 400
+
+    except RequestException:
+        logger.info("AJV Schema validator service unavailable")
+        return json.dumps(obj={}, error="AJV Schema validator service unavailable")
+
+    validator = QuestionnaireValidator(json_to_validate)
+    validator.validate()
+
+    if len(validator.errors) > 0:
```
liamtoozer, line 82: V minor suggestion - This was already here before, but could do `if validator.errors` instead?

api.py
```diff
@@ -1,11 +1,97 @@
...
+
+    return response
 
-application = Flask(__name__)
-application.register_blueprint(validate_blueprint)
-application.register_blueprint(status_blueprint)
 
 if __name__ == "__main__":
-    application.run(port=5001)
+    uvicorn.run("api:app", workers=20, port=5001, reload=True)
```
liamtoozer, line 97: I get the below warning message when I run `make run`, due of the `reload` flag I think? 

```
WARNING:  "workers" flag is ignored when reloading is enabled.
```

liamtoozer (commented):
Great work! Tested locally

api.py
petechd: So I guess I will remove it, the "reload".

petechd: Yeah , it's only for our local runs, so I guess we don't want multiple workers as much as reloads. It is easily adjustable in runner and validator anyway, just depends what do we want more? Fast validation or server auto reloading on changes.

petechd: I guess we only should validate the schemas affected when testing locally so the speed of extra workers is not essential, I will keep reloads on.

liamtoozer: Yeah that makes sense 👍  My vote goes to 'reloads' but see what others think?

Dockerfile
```diff
@@ -17,7 +17,8 @@ RUN poetry install --no-dev
 
 EXPOSE 5000
 
-ENV FLASK_APP=api.py
-
-ENTRYPOINT flask run --host 0.0.0.0
-
+CMD ["gunicorn", "api:app", \
+     "--bind", "0.0.0.0:5000", \
+     "--workers", "20", \
```
berroar, line 22: Is there any documentation as part of this spike around how we ended up at 20 workers?

api.py
petechd: Not sure if this will work as the `else data` replaces a line where it's checked if it's a dict in the original change.

api.py
petechd: Added.

api.py
```diff
@@ -1,11 +1,97 @@
...
+AJV_VALIDATOR_URL = f"http://{AJV_HOST}:5002/validate"
+
+
+@app.post("/validate")
+async def validate_schema_request_body(payload=Body(None)):
+    logger.info("Validating schema")
+    return await validate_schema(payload)
+
+
+@app.get("/validate")
```
berroar, line 35: We should probably document this in the readme, how this works, as not sure it recorded anywhere. Not necessarily as part of this change though 👍

api.py
```diff
@@ -1,11 +1,97 @@
...
+async def validate_schema_from_url(url: str = ""):
```
berroar: Should it be None rather than empty string here?

api.py
```diff
@@ -1,11 +1,97 @@
-from flask import Flask
+import json
+import os
+import urllib
```
berroar, line 3: Minor but maybe just import the actual methods rather than the whole lib?

Dockerfile
petechd: Done this plus expanded on the "-b" attribute as well to make it more readable.

api.py
berroar: Worth adding type hints for this repo (and others) in their own tickets, so we can ensure they are progressed and reviewed properly. But def worth doing 👍

api.py
liamtoozer: I'm not 100% sure but if we're only parsing a str or dict then it might be okay? Looks like it was intentionally checking for dict explicitly before though like you said, so happy to leave it as it was 👍

Dockerfile
petechd: I mean I can add something somewhere but I don't have the exact timings for the experimental settings anymore, I like multithreaded workers, less workers, less parallel shells etc. But I can list what I've tried.

api.py
berroar: I would call this out in readme, that if you want it running fast/workers locally need to set reload to False.

Dockerfile
berroar: Yeah that would be great 👍 Just need some rationale really

api.py
petechd: I can try writing something, better than having nothing like now.

api.py
petechd: Added.

api.py
petechd: Added to the README now, thanks.

berroar (approved):

liamtoozer (approved):

PR #253, opened 2024-09-19
Update to Python 3.12 & dependencies

liamtoozer:
We need to upgrade to Python 3.12 and bump our dependencies to ensure we are up-to-date with important features, security patches etc. In addition, extra bits added were:

* Added `pip install setuptools` as a workaround to Dockerfile - `setuptools` has been deprecated and is no longer bundled with cpython as of 3.12, and needs to be installed separately
* Dropped the `pylint-quotes` dependency - we can now switch over to using `flake8-quotes` due to its ongoing maintenance and support (as we do already in Runner)
* Replaced `--no-dev` with `--only main` in from poetry install command

### How to review
Ensure the build passes and that you're able to run validator locally and with Docker

### Checklist


petechd (approved):
Tested, works great!

berroar (approved):

PR #260, opened 2025-01-13
Remove Individual Response

Yuyuutsu:
Remove Individual Response (IR) Feature from Validator 

### How to review
Check if Validator works as expected without Individual Response. 
Check if Changes make sense and tests pass as expected
Check if all references to IR have been removed.
### Checklist


petechd (approved):

VirajP1002 (approved):
LGTM! 🎉

Yuyuutsu:
Given the uncertainty around Social Surveys and Census we need to park this until we understand whether the current Individual Response feature is needed.

PR #262, opened 2025-01-22
Add Permissions to Workflows, Pin Unpinned Tags for Non-Immutable Actions, and Enforce Full Server-Side Forgery Checks

Yuyuutsu:
Code Scanning has been enabled in some of our Github Repositories is highlighting issues. 
This PR resolves these issues:

- Workflow does not contain permissions
- Unpinned tag for a non-immutable Action in workflow
- Full Server Side Forgery

### How to review
Check changes resolve CodeQL Scanning Errors
Test the URL validation changes

### Checklist


berroar (changes requested):
Had a more in depth look at the issue flagged for: [Full server-side request forgery](https://github.com/ONSdigital/eq-questionnaire-validator/security/code-scanning/8), and I do think we should be handling this. 

I know we spoke about it at tech session previously, but I didn't realise that it only applies to the `validate_schema_from_url` method/endpoint. So I think we should add some logic to ensure that any schema being validated either comes from an ONSDigital repo or a Github gist? I'm not sure there are any other use cases for this that would mean that we need to add any other domains, but would resolve this for the time being.

berroar (changes requested):
Attempted to validate a schema from our schemas repo at the moment, using the validate from url method, and this doesn't work at the moment? Also need to update the readme once this is working, to detail this new rule

Yuyuutsu:
> Attempted to validate a schema from our schemas repo at the moment, using the validate from url method, and this doesn't work at the moment? Also need to update the readme once this is working, to detail this new rule

I have updated so it also allows the raw.githubusercontent.com domain now. Tested and works as expected. Also updated the read me.

.application-version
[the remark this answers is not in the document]
Yuyuutsu: Removed 👍

.github/workflows/release.yml
```diff
@@ -4,13 +4,16 @@ on:
   release:
     types: [published]
 
+permissions:
+  contents: read
```
petechd, line 8: Remember we had this problem with schemas repo where we needed write access to create a zip file to download, don't we need that kind of access here as well?

api.py
```diff
@@ -36,6 +37,19 @@ async def validate_schema_request_body(payload=Body(None)):
 async def validate_schema_from_url(url=None):
```
berroar, line 37: This is not working for me at the moment? Getting the 400 error when I try to validate a schema from the questionnaire schemas repo? Is this working for you?

api.py
```diff
@@ -36,6 +37,19 @@ async def validate_schema_request_body(payload=Body(None)):
...
     if url:
         logger.info("Validating schema from URL", url=url)
+        parsed_url = urlparse(url)
+        if not is_hostname_allowed(parsed_url.hostname):
+            return Response(
+                status_code=400,
+                content=f"URL domain [{parsed_url.hostname}] is not allowed",
+            )
+        if parsed_url.hostname == "github.com" and not parsed_url.path.startswith(
```
petechd, line 46: You already do this check in `is_hostname_allowed` (checking if hostname is "github.com") so might want as well just use that function here and then do `startswith()`.
```python
if is_hostname_allowed(parsed_url.hostname) and not parsed_url.path.startswith(
```
could as well just do else since if "not allowed" condition is `false` it will be allowed, so:
```python
if not is_hostname_allowed(parsed_url.hostname):
            return Response(
                status_code=400,
                content=f"URL domain [{parsed_url.hostname}] is not allowed",
            )
elif not parsed_url.path.startswith("/ONSdigital/"
        ):
```
or for cleanliness invert the whole thing to avoid "not":
```python
if is_hostname_allowed(parsed_url.hostname) and not parsed_url.path.startswith("/ONSdigital/"
        ):
        ....
elif not is_hostname_allowed(parsed_url.hostname):
```
To avoid repeating yourself you could do the ` is_hostname_allowed(parsed_url.hostname)` once or assign value of `parsed_url.hostname` to a var.

berroar, line 46: I don't think it is possible to validate the schema from github without it being the raw json or a gist? So not sure this if ever applies?
```
if parsed_url.hostname == "github.com" and not parsed_url.path.startswith(
    "/ONSdigital/"
):
```

.github/workflows/release.yml
Yuyuutsu: Updated to have write permissions too 👍

api.py
Yuyuutsu: No it cannot, I have updated it so it does not have the github.com domain 👍

api.py
Yuyuutsu: I have updated the logic now so it only uses is host name allowed now. I have removed the github.com check since it is not needed anymore as we cannot validate via url with a github.com domain. 
I have also changed the domains from a list to a set.

Yuyuutsu: I have removed the github.com check since it is not needed anymore as we cannot validate via url with a github.com domain.
I have also changed the domains from a list to a set.
🫡

petechd (commented):
Tested (validate by url) and it works fine now, the release.yml might not actually need the write permissions, as we don't create the .zip file like in schemas repo. Tested it in a clone with only read permissions and it was fine.

api.py
```diff
@@ -88,5 +96,18 @@ async def validate_schema(data):
     return response
 
 
+def is_hostname_allowed(hostname):
```
petechd, line 99: You could as well just make it one-liner:
```python
def is_hostname_allowed(hostname):
    allowed_full_domains = {
        "gist.githubusercontent.com",
        "raw.githubusercontent.com",
    }
    allowed_base_domains = {"onsdigital.uk"}

    return hostname in allowed_full_domains or any(
        hostname.endswith(f".{base}") for base in allowed_base_domains
    )
```
so if `True` or `True`

petechd: This seems more readable as well.

berroar (approved):

PR #264, opened 2025-03-17
Make the AJV Validator URL configurable

Yuyuutsu:

### How to review

### Checklist

### DIFF
--- a/api.py
+++ b/api.py
@@ -22,9 +22,7 @@ async def status():
 
 logger = get_logger()
 
-AJV_HOST = os.getenv("AJV_HOST", "localhost")
-
-AJV_VALIDATOR_URL = f"http://{AJV_HOST}:5002/validate"
+AJV_VALIDATOR_URL = os.getenv("AJV_VALIDATOR_URL", "http://localhost:5002/validate")
 
 
 @app.post("/validate")


api.py
```diff
@@ -22,9 +22,7 @@ async def status():
 
 logger = get_logger()
 
-AJV_HOST = os.getenv("AJV_HOST", "localhost")
-
-AJV_VALIDATOR_URL = f"http://{AJV_HOST}:5002/validate"
+AJV_VALIDATOR_URL = os.getenv("AJV_VALIDATOR_URL", "http://localhost:5002/validate")
```
petechd, line 25: I know we didn't have this documented before but this seems like our oversight, since we only have this single env var in Validator. Since we want to "update any documentation/Readme with the above" as part of this task maybe it is worth documenting this env var in the `## Running` part of the README?

Yuyuutsu: Yeah i was wondering where to add it. I have added a section under `Running`. Let me know if it needs any changes 👍

berroar (changes requested):
Has this been tested with the services that call Validator? I don't think it will work at the moment as they will be referencing the old env var

api.py
berroar, line 25: I think there is potential now to create a url that will break validator, in that you need to make sure the Port is 5002 since this is hardcoded in the app. So we probably need to handle this in some way.

martyncolmer: There won't be a port configuration when deployed in GCP as the backend service is referenced directly.

berroar: But guess we still need to handle the local use case? For our dev and PR wokflows etc

Yuyuutsu: I have currently changed it set to: 
`AJV_VALIDATOR_URL = f"{AJV_VALIDATOR_SCHEME}://{AJV_VALIDATOR_HOST}:5002/validate"`

api.py
```diff
@@ -22,9 +22,13 @@ async def status():
...
 logger = get_logger()
 
-AJV_HOST = os.getenv("AJV_HOST", "localhost")
+AJV_VALIDATOR_SCHEME = os.getenv("AJV_VALIDATOR_SCHEME", "http")
+AJV_VALIDATOR_HOST = os.getenv("AJV_VALIDATOR_HOST", "localhost")
+AJV_VALIDATOR_PORT = os.getenv("AJV_VALIDATOR_PORT", "5002")
 
-AJV_VALIDATOR_URL = f"http://{AJV_HOST}:5002/validate"
+AJV_VALIDATOR_URL = (
+    f"{AJV_VALIDATOR_SCHEME}://{AJV_VALIDATOR_HOST}:{AJV_VALIDATOR_PORT}/validate"
```
martyncolmer, line 30: Can we make it -
f"{AJV_VALIDATOR_SCHEME}://{AJV_VALIDATOR_HOST}/validate"
When deploying in GCP we won't be using port at all as the load balancer will redirect to the service.

api.py
```diff
@@ -22,9 +22,16 @@ async def status():
 
 logger = get_logger()
 
-AJV_HOST = os.getenv("AJV_HOST", "localhost")
+AJV_VALIDATOR_SCHEME = os.getenv("AJV_VALIDATOR_SCHEME", "http")
```
martyncolmer, line 25: How about this?

AJV_VALIDATOR_SCHEME = os.getenv("AJV_VALIDATOR_SCHEME", "http")
AJV_VALIDATOR_HOST = os.getenv("AJV_VALIDATOR_HOST", "localhost")
AJV_VALIDATOR_PORT = os.getenv("AJV_VALIDATOR_PORT", "5002")
AJV_VALIDATOR_URL  = os.getenv("AJV_VALIDATOR_URL", "")

if AJV_VALIDATOR_URL== "":
    AJV_VALIDATOR_URL = (
        f"{AJV_VALIDATOR_SCHEME}://{AJV_VALIDATOR_HOST}:{AJV_VALIDATOR_PORT}/validate"
    )
else:
    AJV_VALIDATOR_URL = f"{AJV_VALIDATOR_URL}/validate"

Yuyuutsu: Yeah happy with this logic. I shall push it now 👍

Yuyuutsu:
> Has this been tested with the services that call Validator? I don't think it will work at the moment as they will be referencing the old env var

I have tested it with Runner now by pointing it to this PR. Validates as expected. Not sure if this is what you meant 😅

berroar:
> > Has this been tested with the services that call Validator? I don't think it will work at the moment as they will be referencing the old env var
> 
> I have tested and rasied a PR with Runner pointing it to this PR. Validates as expected. Not sure if this is what you meant 😅

What I mean is we are still referring to the old env var in other repos, and before you defaulted the value it was broken.

Yuyuutsu:
> > > Has this been tested with the services that call Validator? I don't think it will work at the moment as they will be referencing the old env var
> > 
> > 
> > I have tested and rasied a PR with Runner pointing it to this PR. Validates as expected. Not sure if this is what you meant 😅
> 
> What I mean is we are still referring to the old env var in other repos, and before you defaulted the value it was broken. Does it work when pointing to this branch in runner actions?

Ahh okay, it took me so long to figure out what was wrong because i kept miss reading the environment variables 🤦‍♂️ . But now validator does work for runner. I will create some more PR's for other repos that use validator

package.json
```diff
@@ -11,8 +11,8 @@
   "scripts": {
     "test": "mocha --require @babel/register ajv/tests/*.js --exit",
     "lint": "npm run lint:schemas && npm run lint:test-schemas && npm run lint:ajv",
-    "start": "DEBUG=ajv-schema-validator node ajv/app.js &",
-    "stop": "kill -9 $(lsof -ti:5002)",
+    "start": "source .env && DEBUG=ajv-schema-validator node ajv/app.js &",
```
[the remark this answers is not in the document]
Yuyuutsu: I have reverted the `start-ajv` command so it does not require a .env file, it should work as expected.
I have also updated the `make stop-ajv` file to be able to pass a PORT environment variable, which is defaulted to 5002

Makefile
```diff
@@ -1,10 +1,11 @@
 .PHONY: build run lint test
+PORT ?= 5002
```
petechd, line 2: It might be not the best practice to set this env var in the Makefile. How do we do this in Runner?

Makefile
```diff
@@ -1,10 +1,11 @@
...
 
 build:
 	poetry run ./scripts/build.sh
 
 stop-ajv:
-	npm run stop
+	@PORT=$(PORT) npm run stop
 
 start-ajv:
 	npm run start
```
petechd, line 11: `start-ajv` also needs the PORT var?

Yuyuutsu: START AJV runs app.js which fetches the PORT from the `.env` file. Unlike STOP AJV which has to hard code the port  it wants to stop

package.json
```diff
@@ -12,7 +12,7 @@
     "test": "mocha --require @babel/register ajv/tests/*.js --exit",
     "lint": "npm run lint:schemas && npm run lint:test-schemas && npm run lint:ajv",
     "start": "DEBUG=ajv-schema-validator node ajv/app.js &",
-    "stop": "kill -9 $(lsof -ti:5002)",
+    "stop": "kill -9 $(lsof -ti:$PORT)",
```
berroar, line 15: I get an error when running this: `lsof: unacceptable port specification in: -i :`

Makefile
```diff
@@ -4,7 +4,7 @@ build:
 	poetry run ./scripts/build.sh
 
 stop-ajv:
-	npm run stop
+	@PORT=$(PORT) npm run stop
```
petechd, line 7: We could avoid passing any vars manually if we follow what we do in Runner, eg:
```
SCHEMAS_VERSION=`cat .schemas-version`
DESIGN_SYSTEM_VERSION=`cat .design-system-version`
RUNNER_ENV_FILE?=.development.env
SCHEMA_PATH=./schemas/test/en/
```
PORT being assigned similar to "SCHEMAS_VERSION=`cat .schemas-version`"

Makefile
petechd: Check my other comment: https://github.com/ONSdigital/eq-questionnaire-validator/pull/264#discussion_r2063054252

Makefile
Yuyuutsu: Are we okay with having a file like .schema-version for the Port number soley for the make file command? If a developer did want to change the PORT number they would have to updaet their `.env `as well as the` .port `file.

Yuyuutsu: I have made the changes but you may experience a similar issue you had before:
	> source .env && DEBUG=ajv-schema-validator node ajv/app.js &
	sh: .env: No such file or directory

berroar (approved):

petechd (commented):
I will wait with approving the schemas PR until you merge this and you change the branch there.

PR #265, opened 2025-03-18
Assign Server Side Forgery Checks and Assign Permissions to Github Actions

Yuyuutsu:
This PR is a follow up to the merged PR: https://github.com/ONSdigital/eq-questionnaire-validator/pull/262

The original PR was meant to resolve the all codeQL scan checks.

However the Server Side Forgery check was still alerted due to an oversight on the URL requiring "https://"

This PR is here to resolve the issue and add it to the list of allowed domains.

### How to review

Confirm that this change resolves the outstanding CodeQL issues:
https://github.com/ONSdigital/eq-questionnaire-validator/security/code-scanning/

I've forked the Validator and tested the fix in my fork:
[Forked Repo CodeQL Scan](https://github.com/Yuyuutsu/yuyu-eq-questionnaire-validator/security/code-scanning)
All related CodeQL issues have been resolved there, which indicates the solution is effective.

Ensure that the Validator still functions as expected after this change.
 

### Checklist

### DIFF
--- a/api.py
+++ b/api.py
@@ -36,9 +36,11 @@ app = FastAPI()
 
 logger = get_logger()
 
+
 @app.get("/status")
 async def status():
     return Response(status_code=200)
+
 
 @app.post("/validate")
 async def validate_schema_request_body(payload=Body(None)):


api.py
```diff
@@ -96,16 +105,19 @@ async def validate_schema(data):
...
-
-    return hostname in allowed_full_domains or any(
-        hostname.endswith(f".{base}") for base in allowed_base_domains
+def is_domain_allowed(parsed_url):
+    base_url = f"{parsed_url.scheme}://{parsed_url.netloc}/"
+    repo_owner = (
+        parsed_url.path.split("/")[1] if len(parsed_url.path.split("/")) > 1 else ""
     )
+    domain = parsed_url.netloc
+    if base_url in ALLOWED_FULL_DOMAINS and repo_owner in ALLOWED_REPO_OWNERS:
```
petechd, line 114: Could be simplified so only one `return`:
```python
return (base_url in ALLOWED_FULL_DOMAINS and repo_owner in ALLOWED_REPO_OWNERS) or domain in ALLOWED_BASE_DOMAINS
```

api.py
```diff
@@ -96,16 +105,19 @@ async def validate_schema(data):
...
-    allowed_base_domains = {"onsdigital.uk"}
-
-    return hostname in allowed_full_domains or any(
-        hostname.endswith(f".{base}") for base in allowed_base_domains
+def is_domain_allowed(parsed_url):
+    base_url = f"{parsed_url.scheme}://{parsed_url.netloc}/"
+    repo_owner = (
+        parsed_url.path.split("/")[1] if len(parsed_url.path.split("/")) > 1 else ""
     )
+    domain = parsed_url.netloc
```
petechd, line 113: Should be created only once above and called twice.

Yuyuutsu: Updated 👍

api.py
```diff
@@ -14,6 +14,15 @@
 
 app = FastAPI()
 
+ALLOWED_FULL_DOMAINS = {
```
liamtoozer, line 17: If we're adding more constants here, it might be worth logically grouping these together with the existing consts, and placing them all near the top of the module?

Yuyuutsu: Good Point. I have pushed some changes, reformatting the constants.

liamtoozer (approved):

PR #272, opened 2025-04-25
Upgrade Node Version to v22.15.0

Yuyuutsu:
Validator currently uses Node v20 which will soon be going under maintenance support only. This PR upgrades our node version to the latest stable release - v22.15.0

### How to review
Ensure new npm commands work as previously
Ensure AJV Validator runs and works as expected

### Checklist


rmccar:
I think we should only update to v22. Only the even numbered versions of node go into LTS and the odd versions become unsupported after 6 months

berroar (approved):

Farhanam76 (approved):

PR #277, opened 2025-06-10
Add Localhost to Allowed Base Domains 

Yuyuutsu:
Validator is not working in Runner locally because localhost is not part of the allowed base domains in validator. This PR adds `localhost` to the allowed domains dictionary

### How to review
Run make run validator in Runner pointing to the tag `add-local-host-domain-to-allowed-list`
Test to see `make validate-test-schemas `works now

### Checklist

### DIFF
--- a/api.py
+++ b/api.py
@@ -17,7 +17,7 @@ ALLOWED_FULL_DOMAINS = {
     "https://raw.githubusercontent.com/",
 }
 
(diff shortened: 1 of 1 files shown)


VirajP1002 (approved):
Works for me!

berroar (approved):

PR #284, opened 2025-08-08
EQS-94 Replace deprecated jsonpath_rw libraries

petechd:
This replaces the usage of the `jsonpath-rw` library with the `jsonpath-ng` library for JSONPath parsing and matching within the QuestionnaireSchema class and related functions. It modernises the codebase by adopting a more actively maintained library and improve maintainability and performance.

Changes in detail:
- Replaced jsonpath-rw's jp.match with jsonpath-ng.ext's ext_parse and .find() for JSONPath operations.
- Updated methods like `get_blocks`, `_get_path_id`, and others to use ext_parse for JSONPath parsing.
- Adjusted logic to extract values from matches using `match.value` instead of directly returning results from jp.match.
- Addressed issues where `DatumInContext` objects were incorrectly accessed, ensuring proper handling by using `.value` attributes.

### How to review
Focus should be on checking if all the logic is still used after the refactor, if we don't omit some parts of the validator. This could be hard to detect so debugging is probably needed, ideally you would like to run two validations alongside each other, one for main and one for these changes.

### Checklist


Yuyuutsu (approved):

app/validators/questionnaire_schema.py
```diff
@@ -481,7 +497,10 @@ def get_first_answer_in_block(self, block_id):
 
     @lru_cache
     def _get_path_id(self, path):
```
VirajP1002, line 499: Quick question, is this function still used? I couldn't see it being used anywhere

petechd: Removed as not being used anymore. Thanks.

VirajP1002 (approved):

PR #286, opened 2025-08-15
EQS 232 - Add logging for Validator

farres1:
### Motivation and Context
Adds additional logging to the Validator project to improve the process of debugging the application.

### What has changed
Additional logs added to the `api.py` file.
Also adds code to handle validation of JSON data received as bytes.

### How to test?
- Start the process of launching and debugging the Validator application
- Ensure the AJV Validator service is also running
- Make requests to Validator (e.g. using Postman)
    - Both the `validate_schema_from_url` and `validate_schema` functions can be reviewed
        - To test the `validate_schema_from_url` function, make a `GET` request to `localhost:5001/validate?url=<URL>` 
        - To test the `validate_schema` function, make a `POST` request to `localhost:5001/validate`
            - Valid example schemas can be found from the `eq-questionnaire-runner` project's `schemas/test` directory
            - Also review questionnaire schemas that would not pass Validator checks (e.g. consecutive spaces)
        - These will either return an empty object `{}` if the schema contains no errors, or an object containing the schema's validation errors otherwise

### DIFF
--- a/api.py
+++ b/api.py
@@ -55,7 +55,10 @@ async def validate_schema_from_url(url=None):
         parsed_url = urlparse(url)
         domain = parsed_url.netloc
         if not is_domain_allowed(parsed_url, domain):
-            logger.warning("Schema validation request rejected - URL not allowed (this could be related to domain, base_url or repo_owner - see debug logs for values)", url=url)
+            logger.warning(
+                "Schema validation request rejected - URL not allowed (this could be related to domain, base_url or repo_owner - see debug logs for values)",
+                url=url,
+            )
             return Response(
                 status_code=400,
                 content=f"URL domain [{parsed_url.hostname}] is not allowed",
@@ -70,7 +73,8 @@ async def validate_schema_from_url(url=None):
                 "Could not load schema from allowed domain - URL not found", url=url
             )
             return Response(
-                status_code=404, content=f"Could not load schema at allowed domain - URL not found [{url}]"
+                status_code=404,
+                content=f"Could not load schema from allowed domain - URL not found [{url}]",
             )
 
 
@@ -86,7 +90,9 @@ async def validate_schema(data):
                 logger.info("JSON data decoded as UTF-8 successfully")
             except UnicodeDecodeError:
                 logger.error("Failed to decode JSON data as UTF-8", status=400)
-                return Response(status_code=400, content="Failed to decode JSON data as UTF-8")            
+                return Response(
+                    status_code=400, content="Failed to decode JSON data as UTF-8"
+                )
         if isinstance(data, str):
             logger.info("JSON data received as string - parsing required")
             logger.debug("Attempting to parse JSON data...")
@@ -102,13 +108,20 @@ async def validate_schema(data):
 
     response = {}
     try:
-        logger.debug("Sending JSON data to AJV Schema Validator service...", url=AJV_VALIDATOR_URL)
+        logger.debug(
+            "Sending JSON data to AJV Schema Validator service...",
+            url=AJV_VALIDATOR_URL,
+        )
         ajv_response = requests.post(
             AJV_VALIDATOR_URL, json=json_to_validate, timeout=10
         )
         if ajv_response_dict := ajv_response.json():
             response["errors"] = ajv_response_dict["errors"]
-            logger.warning("AJV Schema Validator service returned errors", status=400, errors=response["errors"])
+            logger.warning(
+                "AJV Schema Validator service returned errors",
+                status=400,
+                errors=response["errors"],
+            )
             return response, 400
 
     except RequestException:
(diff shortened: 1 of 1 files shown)


api.py
```diff
@@ -44,83 +44,146 @@ async def status():
...
                 status_code=400,
                 content=f"URL domain [{parsed_url.hostname}] is not allowed",
             )
-
+        logger.info("Schema validation request accepted - URL allowed", url=url)
         try:
+            # Opens the URL and validates the schema
             with request.urlopen(parsed_url.geturl()) as opened_url:
                 return await validate_schema(data=opened_url.read().decode())
+            logger.info("Schema successfully validated from URL", url=url)
```
Yuyuutsu, line 71: I think this logger is unreachable as it is after the return

farres1: This line was unreachable code. I have now removed this `info` log - if this log were to be output, it would be repeating information previously output in other logs, since the `validate_schema` function that would run before it includes a similar log.

api.py
```diff
@@ -44,83 +44,145 @@ async def status():
...
-        if isinstance(data, str):
-            try:
-                json_to_validate = json.loads(data)
-            except JSONDecodeError:
-                logger.info("Could not parse JSON", status=400)
-                return Response(status_code=400, content="Could not parse JSON")
-        elif isinstance(data, dict):
+        # Sets `json_to_validate` to the data received if it is a dictionary as data does not require processing
+        if isinstance(data, dict):
+            logger.info("JSON data received as dictionary - processing not required")
```
petechd, line 86: I'm running it locally now and the `processing not required` could be a little less ambiguous and slightly misleading (too broad term), can we change it to?:
```suggestion
            logger.info("JSON data received as dictionary - decoding and parsing not required")
```
so it's like:
Untitled 6

api.py
```diff
@@ -44,83 +44,145 @@ async def status():
...
 
 
 def is_domain_allowed(parsed_url, domain):
+    logger.debug("Checking if domain is allowed...", domain=domain)
     base_url = f"{parsed_url.scheme}://{parsed_url.netloc}/"
     repo_owner = (
         parsed_url.path.split("/")[1] if len(parsed_url.path.split("/")) > 1 else ""
     )
+    logger.debug("Parsed URL components", base_url=base_url, repo_owner=repo_owner)
     return (
```
petechd, line 155: We could be a little more explicit here and log what exactly happened so:
Untitled 7
or even:
Untitled 8

api.py
```diff
@@ -44,83 +44,145 @@ async def status():
...
         base_url in ALLOWED_FULL_DOMAINS and repo_owner in ALLOWED_REPO_OWNERS
     ) or domain in ALLOWED_BASE_DOMAINS
 
 
+def decode_and_parse_json(data):
+    processed_data = data
+    # Decodes `data` to string if it is in bytes format
+    if isinstance(data, bytes):
```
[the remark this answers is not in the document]
farres1: Discussion during call:

Currently sending `data` as bytes does not happen in live scenarios - code related to decoding bytes has been removed based on this.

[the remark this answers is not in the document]
farres1: Discussion from call with @petechd:

Currently sending `data` as bytes does not happen in live scenarios - code related to decoding bytes has been removed based on this.

Yuyuutsu (approved):
LGTM!

api.py
```diff
@@ -44,82 +74,171 @@ async def status():
...
-        return json.dumps(obj={}, error="AJV Schema validator service unavailable")
+        logger.error("AJV Schema Validator service unavailable")
+        return json.dumps(obj={}, error="AJV Schema Validator service unavailable")
+
+    logger.info("AJV Schema Validator service returned no errors", status=200)
 
     validator = QuestionnaireValidator(json_to_validate)
+    logger.debug(
+        "Attempting to validate questionnaire schema contents with Questionnaire Validator...",
+        questionnaire_title=json_to_validate.get("title"),
```
petechd, line 160: One last request, can we extract `survey_id` and `form_type` here as well? Title can be slightly misleading sometimes (and repeated for multiple schemas) while with `title` combined with `survey_id`, `form_type` there are no two identical surveys? The only thing is `form_type` is not required field but appears in all business schemas.

petechd: Ideally, we would like schema file name here but that would require a lot of extra work..

configs/logging_config.py
```diff
@@ -0,0 +1,31 @@
+import logging
+import os
+import sys
+
+import structlog
+
+
+def configure_logging():
+    log_level = logging.DEBUG if os.getenv("LOG_LEVEL") == "DEBUG" else logging.INFO
```
petechd, line 9: We should add something in the README about this env var and what it does? Or should @rmccar add it as part of his README update?

api.py
```diff
@@ -44,82 +46,169 @@ async def status():
...
+                status=400,
+                errors=response["errors"],
+            )
             return response, 400
 
     except RequestException:
-        logger.info("AJV Schema validator service unavailable")
-        return json.dumps(obj={}, error="AJV Schema validator service unavailable")
+        logger.exception("AJV Schema Validator service unavailable")
+        return json.dumps(obj={}, error="AJV Schema Validator service unavailable")
```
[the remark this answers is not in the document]
liamtoozer: Can see this has been fixed in 74e901bde8f5e2952bb0087045ae4a0c37f25ab6

PR #380, opened 2026-04-09
Manual package updates

Paul-Joel:
### Motivation and Context

Manual update of updatable npm packages

### What has changed

Bump eslint-plugin-chai-friendly to 1.2.0
Bump prettier to 3.8.1
Bump Globals to 17.5.0
Tweaks to tox.ini and run_lint_python to prevent local checks of package files

### How to test?

Check tests pass etc

### DIFF
--- a/scripts/run_lint_python.sh
+++ b/scripts/run_lint_python.sh
@@ -18,22 +18,31 @@ function display_result {
     fi
 }
 
-flake8 --max-complexity 10 --count
+COMMON_EXCLUDES=".venv,.tox,node_modules,htmlcov,megalinter-reports"
+BLACK_EXCLUDES='/(\.venv|\.tox|node_modules|htmlcov|megalinter-reports)/'
+
+flake8 --max-complexity 10 --count --exclude "$COMMON_EXCLUDES"
 display_result $? 1 "Flake 8 code style check"
 
-find . -type f -name "*.py" | xargs pylint --reports=n --output-format=colorized --rcfile=.pylintrc -j 0
+find . \
+    \( -path "./.venv" -o \
+    -path "./.tox" -o \
+    -path "./node_modules" -o \
+    -path "./htmlcov" -o \
+    -path "./megalinter-reports" \) -prune \
+    -o -type f -name "*.py" -print | xargs pylint --reports=n --output-format=colorized --rcfile=.pylintrc -j 0
 # pylint bit encodes the exit code to allow you to figure out which category has failed.
 # https://docs.pylint.org/en/1.6.0/run.html#exit-codes
 # We want to fail on all errors so don't check for specific bits in the output; but if we did in future, see:
 # http://stackoverflow.com/questions/6626351/how-to-extract-bits-from-return-code-number-in-bash
 display_result $? 2 "Pylint linting check"
 
-isort --check .
+isort --check --skip .venv --skip .tox --skip node_modules --skip htmlcov --skip megalinter-reports .
 display_result $? 1 "isort linting check"
 
-black --check . --exclude node_modules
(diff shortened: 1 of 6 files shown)


scripts/run_lint_python.sh
```diff
@@ -18,22 +18,31 @@ function display_result {
...
+    -path "./htmlcov" -o \
+    -path "./megalinter-reports" \) -prune \
+    -o -type f -name "*.py" -print | xargs pylint --reports=n --output-format=colorized --rcfile=.pylintrc -j 0
 # pylint bit encodes the exit code to allow you to figure out which category has failed.
 # https://docs.pylint.org/en/1.6.0/run.html#exit-codes
 # We want to fail on all errors so don't check for specific bits in the output; but if we did in future, see:
 # http://stackoverflow.com/questions/6626351/how-to-extract-bits-from-return-code-number-in-bash
 display_result $? 2 "Pylint linting check"
 
-isort --check .
```
petechd, line 31: Not sure if we don't want to use the preferred method according to isort docs:

Image

scripts/run_lint_python.sh
```diff
@@ -18,22 +18,31 @@ function display_result {
...
+isort --check --skip .venv --skip .tox --skip node_modules --skip htmlcov --skip megalinter-reports .
 display_result $? 1 "isort linting check"
 
-black --check . --exclude node_modules
```
petechd, line 34: For black can we take advantage of this?

Image

scripts/run_lint_python.sh
```diff
@@ -18,22 +18,31 @@ function display_result {
     fi
 }
 
-flake8 --max-complexity 10 --count
+COMMON_EXCLUDES=".venv,.tox,node_modules,htmlcov,megalinter-reports"
+BLACK_EXCLUDES='/(\.venv|\.tox|node_modules|htmlcov|megalinter-reports)/'
+
+flake8 --max-complexity 10 --count --exclude "$COMMON_EXCLUDES"
```
petechd: And for flake8 use config in pyproject.tom?

Image

in pyproject.toml it would be slightly different, an example:
```
[tool.flake8]
# Check that this is aligned with your other tools like Black
max-line-length = 120
exclude = [
    # No need to traverse our git directory
    ".git",
    # There's no value in checking cache directories
    "__pycache__"
]
# Use extend-ignore to add to already ignored checks which are anti-patterns like W503.
extend-ignore = [
    # PEP 8 recommends to treat : in slices as a binary operator with the lowest priority, and to leave an equal
    # amount of space on either side, except if a parameter is omitted (e.g. ham[1 + 1 :]).
    # This behaviour may raise E203 whitespace before ':' warnings in style guide enforcement tools like Flake8.
    # Since E203 is not PEP 8 compliant, we tell Flake8 to ignore this warning.
    # https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#slices    
    "E203"
]
```

petechd:
For ease of browsing the PRs later, can we mention the updated packages in the PR title?

scripts/run_lint_python.sh
Paul-Joel: Removed the --exclude, and works fine

scripts/run_lint_python.sh
Paul-Joel: Have removed the changes as not actually required.

scripts/run_lint_python.sh
Paul-Joel: Implemented as suggested. Underlying issue solved fine with changes

petechd (approved):
If you could only mention packages names in the PR title it would be great.

VirajP1002 (approved):

PR #387, opened 2026-04-15
Eqs 766 add html validation

oms09:
### Motivation and Context

After the investigation into validator gaps, HTML tags and entities was not being validated.

### What has changed

- Added validate_html() to iterate through all translatable schema text.
- Implemented HTML tag validation using a stack-based approach. 
- Ensures only allowed tags are used (p, strong, a, b, br, img), validates correct opening/closing tag structure, handles self-closing tags (br, img)  
- Added validation for HTML entities Supports numeric (decimal and hexadecimal) and named entities, uses html.entities.html5 for validation
-  Detects whitespace before br tags
- Ensures p tags appear at the start of a sentence (ignoring whitespace and wrapper characters) 
- unit tests added 

### How to test?

make test
run validator and runner then run make validate-test-schemas to ensure there are no errors
check code

### Links

https://officefornationalstatistics.atlassian.net/browse/EQS-766
Rhys Validation PR - https://github.com/ONSdigital/eq-questionnaire-validator/pull/244


app/error_messages.py
```diff
@@ -1,6 +1,9 @@
 """Error messages used in validators and exceptions throughout the validator codebase."""
 
 DUMB_QUOTES_FOUND = "Found dumb quotes(s) in schema text"
+HTML_FOUND = "Found invalid HTML tag(s) in schema text"
```
VirajP1002: Would it be better to prefix this with `INVALID_`? As it implies any HTML is bad 😅

Same would go for `HTML_ENTITIES_FOUND`.

app/validators/questionnaire_validator.py
VirajP1002, line 37: Minor, may be worth updating the Docstring here to list your new methods?

app/validators/questionnaire_validator.py
```diff
@@ -167,6 +170,161 @@ def validate_smart_quotes(self):
...
+        Args:
+        text (str): The text to validate.
+        pointer (str): JSON pointer to the location of the text in the schema.
+        """
+        match = re.search(r"<p(?=[\s>])[^>]*>", text)
+
+        if not match:
+            return
+
+        text_before_p = text[: match.start()]
```
VirajP1002: Minor, could rename this as `text_before_p_tag`? Just for clarity and to be consistent with the method name? Same would go for `content_before_p` --> `content_before_p_tag`

app/validators/questionnaire_validator.py
```diff
@@ -167,6 +170,161 @@ def validate_smart_quotes(self):
...
+                pointer=pointer,
+                text=text,
+            )
+
+    def check_p_tag_position(self, text, pointer):
+        """Checks if p tag is at the start of a sentence
+        (ignoring whitespace and wrapper characters like [] and ()).
+
+        Args:
+        text (str): The text to validate.
```
VirajP1002: Very very minor, but I would indent the docstring here

farres1 (approved):

app/validators/questionnaire_validator.py
```diff
@@ -167,6 +170,161 @@ def validate_smart_quotes(self):
...
+        for entity in entity_matches:
+            if not self.is_valid_html_entity(entity):
+                self.add_error(
+                    error_messages.INVALID_HTML_ENTITIES_FOUND,
+                    pointer=pointer,
+                    text=text,
+                )
+                return
+
+    def check_br_tag_whitespace(self, text, pointer):
```
VirajP1002: Minor, do you think you think we should add something in the method handler that says that it's check whitespace before a `br` tag? Just for clarity? Maybe something like:

`check_whitespace_before_br_tag` or `check_br_tag_whitespace_prefix`

tests/schemas/invalid/test_invalid_html_tags.json
```diff
@@ -0,0 +1,133 @@
...
+              }
+            },
+            {
+              "type": "Interstitial",
+              "id": "interstitial-five",
+              "content": {
+                "title": "Valid double anchor.",
+                "contents": [
+                  {
+                    "description": "<a href='link'>Title</a> and <a href='link-2'>Not valid tag</a>"
```
VirajP1002: When I was debugging the test, no error was raised for this? Or is the text `Not valid tag` incorrect here?

VirajP1002: Ahh right, double anchors are valid, you probably just need to update the text 😆

app/validators/questionnaire_validator.py
```diff
@@ -41,6 +42,7 @@ class QuestionnaireValidator(Validator):
         validate_smart_quotes
         validate_white_spaces
         validate_introduction_block
+        validate_html
```
VirajP1002, line 45: I think you need to add your other new methods in the Docstring? `check_br_tag_whitespace`, `is_valid_html_entity`, `check_html_tags`, etc.

app/validators/questionnaire_validator.py
oms09: updated 
Also, I think whitespace_before_br_tag is alright because its literally that if you have a whitespace before a br tag it messes things up

tests/schemas/invalid/test_invalid_html_tags.json
oms09: oh yeah you're right I copy and pasted the invalid html tags from rhys PR but yeah :+1:

app/validators/questionnaire_validator.py
oms09: added

VirajP1002 (approved):

PR #412, opened 2026-05-19
Manual bump of Axios and Eslint

VirajP1002:
### Motivation and Context

Manual bumps of Axios and Eslint are required as dependabot couldn't resolve some transitive dependencies.

### What has changed

- Bump Axios from 1.16.0 --> 1.16.1
- Bump Eslint from 9.39.4 --> 10.4.0

### How to test?

Ensure Actions pass and dependencies install

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "ISC",
       "dependencies": {
         "ajv": "8.20.0",
-        "axios": "1.16.0",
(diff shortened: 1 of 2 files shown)


VirajP1002:
`eslint` v10 is currently not supported by the `neostandard` package, therefore we cannot upgrade `eslint` to v10. The `neostandard` repository is being actively maintained and they are currently working on providing support for v10.

PR #458, opened 2026-07-24
Upgrade Python to 3.14 latest patch

ems-sato:
### Motivation and Context
Moves from Python 3.13 to 3.14 as is the current maintenance release

### What has changed
Bumped pinned interpreter to 3.14.6 and raised poetry constraint to 3.14. Moved base container image to 3.14

### How to test?
Run the automated test and check that the image builds and runs

### Links
(https://officefornationalstatistics.atlassian.net/browse/EQS-458)


Farhanam76 (approved):

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.4.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.2 and should not be changed by hand.
```
VirajP1002, line 1: Only very minor, I just noticed that we might want to use Poetry `2.4.1` since that's the latest?

VirajP1002: Should stay as `2.1.2` since it hasn't been updated yet 👍

VirajP1002 (approved):

